Issue #18, opened 2017-07-06
Searching network based on EA's not working when body is passed in GET request


ishantt:
The call `https://github.com/infobloxopen/docker-infoblox/blob/v1.1.0-pre/driver/infoblox_ipam.go#L117`
 
`networkByName, err := ibDrv.objMgr.GetNetwork(netview, "", ibclient.EA{"Network Name": networkName})`
 
eventually constructs the URL as 
 ```
$ curl -k -u admin:infoblox -H "Content-Type:application/json" -X GET 'https://10.120.21.150:443/wapi/v2.6/network?%2ANetwork+Name=new-net1&_return_fields=extattrs%2Cnetwork%2Cnetwork_view' -d '{"network_view":"global"}
 
'[
    {
        "_ref": "network/ZG5zLm5ldHdvcmskNDAuMC4wLjAvMjQvMA:40.0.0.0/24/default",
        "extattrs": {
            "CMP Type": {
                "value": "Docker"
            },
            "Cloud API Owned": {
                "value": "True"
            },
            "Network Name": {
                "value": "new-net1"
            },
            "Tenant ID": {
                "value": "DVZR:HNJZ:42OG:XTRO:YOHD:VDYA:EBKK:UDO7:ILEA:JF7R:KYGG:QCIO"
            }
        },
        "network": "40.0.0.0/24",
        "network_view": "default"
    }
 ```
It queried for a network with EA “network-name = new-net1” in “global” network view but got the network with this name in “default” view. The GET request ignores the body if there are parameters in the query string. This request should have failed but currently passes in our code.
 
 So either all parameters needs to be passed in the body or as query parameters. 
 
 
 1.	Passing all prams as body
 ```
$ curl -k -u admin:infoblox -H "Content-Type:application/json" -X GET 'https://10.120.21.150:443/wapi/v2.6/network?_return_fields=extattrs%2Cnetwork%2Cnetwork_view' -d '{ "*Network Name":"new-net1"}'
 
[
    {
        "_ref": "network/ZG5zLm5ldHdvcmskNDAuMC4wLjAvMjQvMA:40.0.0.0/24/default",
        "extattrs": {
            "CMP Type": {
                "value": "Docker"
            },
            "Cloud API Owned": {
                "value": "True"
            },
            "Network Name": {
                "value": "new-net1"
            },
            "Tenant ID": {
[...]

PR #19, opened 2017-07-06
Search EA's by passing in body instead of query string

ishantt:
This patch passes EA's in body instead of query string because of
the issue #18.
Also Fixes issues where error is not returned when there is error
in the request response body.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -183,35 +183,26 @@ func (wrb *WapiRequestBuilder) BuildUrl(t RequestType, objType string, ref strin
 }
 
 func (wrb *WapiRequestBuilder) BuildBody(t RequestType, obj IBObject) []byte {
-	var jsonStr []byte
+	var objJson []byte
 	var err error
 
-	jsonStr, err = json.Marshal(obj)
+	objJson, err = json.Marshal(obj)
 	if err != nil {
-		log.Printf("Cannot marshal payload: '%s'", obj)
+		log.Printf("Cannot marshal object '%s': %s", obj, err)
 		return nil
 	}
 
 	eaSearch := obj.EaSearch()
 	if t == GET && len(eaSearch) > 0 {
-		payload := make(map[string]string)
-		json.Unmarshal(jsonStr, &payload)
-		for k, v := range eaSearch {
-			str, ok := v.(string)
-			if !ok {
-				log.Printf("Cannot marshal EA Search attribute for '%s'\n", k)
-			} else {
-				payload["*"+k] = str
-			}
-		}
-		jsonStr, err = json.Marshal(payload)
+		eaSearchJson, err := json.Marshal(eaSearch)
 		if err != nil {
-			log.Printf("Cannot marshal EA's in the payload: '%s'", payload)
+			log.Printf("Cannot marshal EA Search attributes. '%s'\n", err)
 			return nil
 		}
+		objJson = append(append(objJson[:len(objJson)-1], []byte(",")...), eaSearchJson[1:]...)
 	}
 
-	return jsonStr
+	return objJson
 }
 
 func (wrb *WapiRequestBuilder) BuildRequest(t RequestType, obj IBObject, ref string) (req *http.Request, err error) {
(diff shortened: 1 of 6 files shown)


connector.go
```diff
@@ -202,6 +192,24 @@ func (wrb *WapiRequestBuilder) BuildBody(obj IBObject) []byte {
...
+				payload["*"+k] = str
+			}
+		}
+		jsonStr, err = json.Marshal(payload)
+		if err != nil {
+			log.Printf("Cannot marshal EA's in the payload: '%s'", payload)
+			return nil
+		}
+	}
+
```
yuewko: I assume this is just a WIP....

Instead of doing the Unmashal -> add EASearch -> Marshall, you should consider doing the following:
1) Define a new type for EASearch field
2) Define a custom Marshaller for that
3) Add EASearch field to be part of the main object (instead of being in IBBase)

Then the original BuildBody() method should be able to handle it.

ishantt: Yes this is WIP as I need to fix unit tests, but I opened the PR for early review.

Thanks Ko! Looks great idea to me. Let me try this out.

ishantt: Hi @yuewko , 
 I tried this but it will not work for our use case. 

Problem is that tag for the easearch will also come but we don't want that. We just want the k/v pairs of the easearch. 

For eg in `https://play.golang.org/p/XbC_RQLgO9` , we want the Marshalled json to be 
`{"Foo":"Bar","Name":"macvlan","CIDR":"10.10.0.0/24","Gateway":"10.10.0.1"}`
but the "EA" field is also coming. 
`{"EA":{"Foo":"Bar"},"Name":"macvlan","CIDR":"10.10.0.0/24","Gateway":"10.10.0.1"}`

This is currently not supported in the current encoding/json package. This functionality is proposed for the future release by addition of the 'inline' tag which does what we want.
`https://github.com/golang/go/issues/6213`

yuewko: @ishantt , thank you for digging into it. I just assumed that what I proposed would work - my bad.

Since the golang proposal has been kicked around since 2013, I doubt that it will be adopted anytime soon - it has already been rejected for 1.9.

In the meantime, what do you think about something like this: https://play.golang.org/p/Qay9YPCv6O

Thanks!

ishantt: Thanks Ko!
This looks good to me. I will submit a new commit with this change.

object_manager.go
```diff
@@ -157,7 +160,7 @@ func BuildNetworkFromRef(ref string) *Network {
 	}
 }
 
-func (objMgr *ObjectManager) GetNetwork(netview string, cidr string, ea EA) (*Network, error) {
+func (objMgr *ObjectManager) GetNetwork(netview string, cidr string, eaSearch EASearch) (*Network, error) {
```
ishantt: This changes the function signature thereby breaking current docker ipam plugin. I am not sure if any other project uses it.

yuewko: You shouldn't change the signature.

More importantly, EAS is meant for internal use only. As far as the user is concerned, it is just EA.

Since EAS and EA are basically the same beast, you can just do a type conversion inside the function when you assign that to `network.eaSearch`
i.e.:
`network.eaSearch = EAS(ea)`

connector.go
```diff
@@ -192,35 +182,43 @@ func (wrb *WapiRequestBuilder) BuildUrl(t RequestType, objType string, ref strin
...
 
-	return jsonStr
+	eaSearch := obj.EaSearch()
+	if t == GET && len(eaSearch) > 0 {
+		eaSearchJson, err := json.Marshal(eaSearch)
+		if err != nil {
+			log.Printf("Cannot marshal EA Search attributes. '%s'\n", err)
+			return nil
+		}
+		objJson = append(append(objJson[:len(objJson)-1], []byte(",")...), eaSearchJson[1:]...)
```
yuewko: nit... and I confess that's how the example I provided was written... but:
`append(objJson[:len(objJson)-1], []byte(",")...)`
can just be:...
`append(objJson[:len(objJson)-1], byte(','))`

ishantt: yeah makes sense. I just blindly copied your example :P

PR #21, opened 2017-07-18
Add Request object and implement IBConnector UpdateObject 

ishantt:
This patch adds support for the 'Request' NIOS object.
IBConnector UpdateObject method is also implemented in this patch.

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -3,6 +3,7 @@ package ibclient
 import (
 	"fmt"
 	"regexp"
+	"encoding/json"
 )
 
 type IBObjectManager interface {
@@ -25,7 +26,7 @@ type IBObjectManager interface {
 type ObjectManager struct {
 	connector IBConnector
 	cmpType   string
-	tenantID  string
+	TenantID  string
 }
 
 func NewObjectManager(connector IBConnector, cmpType string, tenantID string) *ObjectManager {
@@ -33,7 +34,7 @@ func NewObjectManager(connector IBConnector, cmpType string, tenantID string) *O
 
 	objMgr.connector = connector
 	objMgr.cmpType = cmpType
-	objMgr.tenantID = tenantID
+	objMgr.TenantID = tenantID
 
 	return objMgr
 }
@@ -42,7 +43,7 @@ func (objMgr *ObjectManager) getBasicEA(cloudApiOwned Bool) EA {
 	ea := make(EA)
 	ea["Cloud API Owned"] = cloudApiOwned
 	ea["CMP Type"] = objMgr.cmpType
-	ea["Tenant ID"] = objMgr.tenantID
+	ea["Tenant ID"] = objMgr.TenantID
 	return ea
 }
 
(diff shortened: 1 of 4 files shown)


objects.go
```diff
@@ -198,3 +198,37 @@ func (v *EADefListValue) UnmarshalJSON(b []byte) (err error) {
...
+
+type RequestBody struct{
+	Data map[string]interface{} `json:"data,omitempty"`
+	Args map[string]string `json:"args,omitempty"`
+	Method string `json:"method"`
+	Object string `json:"object,omitempty"`
+	EnableSubstitution bool `json:"enable_substitution,omitempty"`
+	AssignState map[string]string `json:"assign_state,omitempty"`
+	Discard bool `json:"discard,omitempty"`
+}
```
yuewko: Is there some restriction on the field values (or combination thereof) based on the request type (i.e. Method="GET" as oppose to Method="PUT")? If so, there should be a factory method for each type of request, e.g. NewGetRequestBody(), with a specific set of arguments..

object_manager.go
```diff
@@ -25,15 +26,15 @@ type IBObjectManager interface {
 type ObjectManager struct {
 	connector IBConnector
 	cmpType   string
-	tenantID  string
+	TenantID  string
```
ishantt: I made TenantID exportable so that I can use the Docker ID in the Lock https://github.com/ishantt/docker-infoblox/blob/swarm/driver/lock.go#L43

object_manager.go
```diff
@@ -129,6 +130,33 @@ func (objMgr *ObjectManager) GetNetworkView(name string) (*NetworkView, error) {
 	return &res[0], nil
 }
 
+func (objMgr *ObjectManager) UpdateNetworkView(ref string, addEA EA, removeEA EA) error {
```
ishantt: 1. Ok I will rename it to UpdateNetworkViewEA . Right now I dont have any need to update the netview view.

2. Yes I did not add it because I did not find it generic

3. Makes sense. I will create a generic method to update EA for the IBObjects and then i will not need a specific method for updating NetworkView EAs. Thanks!

object_manager.go
```diff
@@ -316,3 +344,24 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
 
 	return newEadef, err
 }
+
+
+func (objMgr *ObjectManager) CreateMultiObjectRequest(body []*RequestBody) ([]map[string]interface{}, error){
```
ishantt: Okay I will change the name.

Actually CreateMultiObjectRequest does not implements IBObject interface because its does not have to implement SearchEA() method 

yeah makes sense to pass MultiRequest object. Thanks!

objects.go
ishantt: No there are no such restrictions on the fields. https://10.120.21.150/wapidoc/objects/request.html

objects.go
```diff
@@ -198,3 +198,37 @@ func (v *EADefListValue) UnmarshalJSON(b []byte) (err error) {
...
+
+type Request struct{
+	IBBase      `json:"-"`
+}
+
+type SingleRequest struct{
+	Request
```
ishantt: Ok. Thanks

object_manager.go
yuewko: So, if you agree that the NVLocker belongs to Infoblox-go-client (as we discussed in the docker-infoblox review), then we don't need to export tenantID, yes?

ishantt:
Thanks Ko for Merging this. Yes I will address the Locker changes a new PR along with the docker-infoblox changes. I did not do that in this PR as it was too big a change for this PR.

yuewko:
Totally agreed! Thanks!!

PR #22, opened 2017-07-20
Add Distributed Network View Lock

ishantt:
This patch adds distributed locking on network view
for concurrent access to resources in a network view.

### DIFF
--- a/lock.go
+++ b/lock.go
@@ -1,0 +1,244 @@
+package ibclient
+
+import (
+	"fmt"
+	"github.com/Sirupsen/logrus"
+	"math/rand"
+	"time"
+)
+
+const (
+	timeout     int32  = 60 // in seconds
+	freeLockVal string = "Available"
+)
+
+type Lock interface {
+	Lock() bool
+	UnLock(force bool) bool
+}
+
+type NVLocker struct {
+	name          string
+	objMgr        *ObjectManager
+	LockEA        string
+	LockTimeoutEA string
+}
+
+func (l *NVLocker) createLockRequest() *MultiRequest {
+
+	req := NewMultiRequest(
+		[]*RequestBody{
+			&RequestBody{
+				Method: "GET",
+				Object: "networkview",
+				Data: map[string]interface{}{
+					"name":         l.name,
+					"*" + l.LockEA: freeLockVal,
+				},
+				Args: map[string]string{
+					"_return_fields": "extattrs",
+				},
+				AssignState: map[string]string{
+					"NET_VIEW_REF": "_ref",
+				},
+				Discard: true,
+			},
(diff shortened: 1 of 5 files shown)


lock.go
```diff
@@ -0,0 +1,244 @@
...
+			&RequestBody{
+				Method: "STATE:DISPLAY",
+			},
+		},
+	)
+
+	return req
+}
+
+func (l *NVLocker) Lock() bool {
```
yuewko: Lock() is now used from within GetNVLock(). Does it still need to be exported?

lock.go
```diff
@@ -0,0 +1,244 @@
...
+	timeout     int32  = 60 // in seconds
+	freeLockVal string = "Available"
+)
+
+type Lock interface {
+	Lock() bool
+	UnLock(force bool) bool
+}
+
+type NVLocker struct {
```
yuewko: Just to be consistent, with the naming of the types the package, you might want to spell out NetworkView - i.e. NetworkViewLocker() instead of NVLocker().

lock.go
```diff
@@ -0,0 +1,244 @@
...
+	)
+
+	return req
+}
+
+func (l *NVLocker) createUnlockRequest(force bool) *MultiRequest {
+
+	getData := map[string]interface{}{"name": l.name}
+	if !force {
+		getData["*"+l.LockEA] = l.objMgr.TenantID
```
yuewko: Don't forget about this: https://github.com/infobloxopen/infoblox-go-client/pull/21#discussion_r128240666
;-)

lock.go
ishantt: ok done.

lock.go
ishantt: yeah forgot about this. Thanks !

lock.go
ishantt: This method implements the Lock interface and another implementation of the Lock (eg GridLock) might not implement GetNVLock() kind of method.

yuewko: OK... I ignored the fact that there is a Locker interface... My bad...

In which case, I think the NetworkViewLocker implementation is problematic. The whole point of a interface is so that any implementation of that interface has similar behavior.

Specifically, a piece of code that utilitize the Locker interface would call Lock() method to create the lock, and, likewise, Unlock() to release the lock - irrespective of the implementation. The way the NetworkViewLocker  is implemented, this wouldn't work.

So, I think you need to refactor the NetworkViewLocker so that it meets that expectation.

ishantt: Yes interface should implement Lock() and Unlock(). Initially I implemented like that only but then when I moved the logic or retry in the Lock.go I changed the implementation. Anyways I will fix this.

PR #23, opened 2017-07-27
Added support for UpdateFixedAddress in client

adsri:
UpdateFixedAddress is required to support cni_spec 0.3.1 as with
cni release v0.4.0: https://github.com/containernetworking/cni/releases/tag/v0.4.0
The mac_address of the interface was allocated based on the ip_address on all the
network drivers, for example:
https://github.com/containernetworking/cni/blob/v0.4.0/plugins/main/macvlan/macvlan.go#L153

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -16,7 +16,7 @@ type IBObjectManager interface {
 	GetNetworkContainer(netview string, cidr string) (*NetworkContainer, error)
 	AllocateIP(netview string, cidr string, ipAddr string, macAddress string, vmID string) (*FixedAddress, error)
 	AllocateNetwork(netview string, cidr string, prefixLen uint, name string) (network *Network, err error)
-	UpdateFixedAddress(fixedAddr *FixedAddress, macAddress string, vmID string) (*FixedAddress, error)
+	UpdateFixedAddress(fixedAddrRef string, macAddress string, vmID string) (*FixedAddress, error)
 	GetFixedAddress(netview string, cidr string, ipAddr string, macAddr string) (*FixedAddress, error)
 	ReleaseIP(netview string, cidr string, ipAddr string, macAddr string) (string, error)
 	DeleteNetwork(ref string, netview string) (string, error)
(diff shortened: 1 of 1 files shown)


object_manager.go
```diff
@@ -304,6 +305,40 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
 	return &res[0], nil
 }
 
+func (objMgr *ObjectManager) UpdateFixedAddress(netview string, cidr string, ipAddr string, macAddress string, vmID string) (*FixedAddress, error) {
```
yuewko: I don't think this code would work if user just wanted to update vmID but not macAddress.

adsri: Oh, sorry you are correct. I will fix it.

yuewko: Thanks!

NOTE that, the point is, if the method is designed to update very specific field(s), the name of the method should indicate so. Example is UpdateNetworkViewEA: https://github.com/infobloxopen/infoblox-go-client/blob/master/object_manager.go#L133

Oh, BTW, it looks like UpdateNetworkViewEA is not in the interface definition (https://github.com/infobloxopen/infoblox-go-client/blob/master/object_manager.go#L9) Maybe it should be included?

Thanks again!

adsri: I intended to make it common update function which can be used for all updates.

yuewko: OK.

adsri:
@yuewko Please review the latest changes as per review comments.

object_manager.go
```diff
@@ -304,6 +309,23 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
 	return &res[0], nil
 }
 
+func (objMgr *ObjectManager) UpdateFixedAddress(fixedAddr *FixedAddress, macAddress string, vmID string) (*FixedAddress, error) {
```
yuewko: If `fixedAddr` is the result of `GetFixedAddress()`, then it would have all it's fields populated. However, we might want to ONLY pass fields that are being updated (plus the Ref) to `UpdateObject()` below. Example, do something like:

`    updateFixedAddr := NewFixedAddress(FixedAddress{Ref: fixedAddress.Ref})`

and then use `updateFixedAddr` instead of `fixedAddr` in the code that follows.

An alternative would be to change this method (`UpdateFixedAddress`) to accept a `Ref `instead of `FixAddress `as the 1st argument.

adsri:
@yuewko Done. Please review the latest changes

PR #24, opened 2017-08-30
Add upgradestatus WAPI object

chinmayb:
WIP: Unit testing

### DIFF
--- a/objects.go
+++ b/objects.go
@@ -124,6 +124,20 @@ func NewEADefinition(eadef EADefinition) *EADefinition {
 	return &res
 }
 
+type UserProfile struct {
+	IBBase `json:"-"`
+	Ref    string `json:"_ref,omitempty"`
+	Name   string `json:"name,omitempty"`
+}
+
+func NewUserProfile(userprofile UserProfile) *UserProfile {
+	res := userprofile
+	res.objectType = "userprofile"
+	res.returnFields = []string{"name"}
+
+	return &res
+}
+
 func (ea EA) MarshalJSON() ([]byte, error) {
 	m := make(map[string]interface{})
 	for k, v := range ea {
--- a/connector.go
+++ b/connector.go
@@ -331,6 +331,12 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 	connector.Requestor.Init(connector.TransportConfig)
 
 	res = connector
-
+	// GET UserProfile request is used here to validate connector's basic auth and reachability.
+	var response []UserProfile
+	userprofile := NewUserProfile(UserProfile{})
+	err = connector.GetObject(userprofile, "", &response)
+	if err != nil {
+		log.Printf("Failed to connect to the Grid, err: %s \n", err)
+	}
 	return
 }


chinmayb:
Please review @ishantt

object_manager.go
```diff
@@ -387,3 +387,20 @@ func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]
 
 	return result, nil
 }
+
+/// GetUpgradeStatus returns the status of the grid information of a map
+func (objMgr *ObjectManager) GetUpgradeStatus(type_ string) ([]map[string]interface{}, error) {
```
ishantt: why type_ and not only type?

chinmayb: `type` is go's keyword, to avoid confusion used type_

yuewko: Please use something more self explanatory, e.g.
If `type` refers to type of upgrade, use `upgradeType`
If it refers to type of status, use `statusType`

object_manager.go
```diff
@@ -387,3 +387,20 @@ func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]
...
+	var res []map[string]interface{}
+
+	if type_ == "" {
+		type_ = "GROUP"
+	}
+	upgradeS := NewUpgradeStatus(UpgradeStatus{Type: type_})
```
yuewko: `upgradeStatus` instead of `upgradeS`

objects.go
```diff
@@ -55,6 +55,20 @@ func NewNetworkView(nv NetworkView) *NetworkView {
 	return &res
 }
 
+type UpgradeStatus struct {
+	IBBase `json:"-"`
+	Ref    string `json:"_ref,omitempty"`
+	Type   string `json:"type"`
+}
+
+func NewUpgradeStatus(ustatus UpgradeStatus) *UpgradeStatus {
```
yuewko: `upgradeStatus` instead of `ustatus`

object_manager.go
chinmayb: Done.

object_manager.go
chinmayb: Done.

objects.go
chinmayb: Done.

chinmayb:
@yuewko @ishantt  i'll update the patch fixing the review comments

object_manager.go
```diff
@@ -387,3 +387,20 @@ func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]
 
 	return result, nil
 }
+
+/// GetUpgradeStatus returns the status of the grid information of a map
+func (objMgr *ObjectManager) GetUpgradeStatus(type_ string) ([]map[string]interface{}, error) {
+	var res []map[string]interface{}
+
+	if type_ == "" {
+		type_ = "GROUP"
```
yuewko: According to v2.6 of WAPI doc, the only acceptable value of type is 'GRID' (please see below). Am I missing something?

```
type
    The upgrade status type.
Type
    String.
Valid values are:
    -  GRID
```

object_manager.go
yuewko: According to WAPI doc (see next comment), the type refers to "upgrade status type".  So, I recommend changing it to `statusType`

object_manager.go
chinmayb: Guess its not supported in that version, i'm trying with v2.7 & v2.8.

here's the  latest:

> 
> type
> The type of upper level elements to be requested.
> 
> Type
> 
> String.
> 
> Valid values are:
> GRID
> GROUP
> PNODE
> VNODE
> 

So should I default it to "GRID" ?

object_manager.go
chinmayb: Sure.

object_manager.go
yuewko: Thanks for the info! In this case, I suggest the following:

1) Considering the nature of the call, I wouldn't even provide a default. It should just validate the provided value against an acceptable set. (i.e. An empty string would be an error.)

2) More importantly, since the set of acceptable values dependent on the WAPI version, we need to take that into account. The python version has a `Feature` class (here: [https://github.com/infobloxopen/infoblox-client/blob/master/infoblox_client/feature.py](url)) that is designed to support this sort of situation. Please take a look and consider implementing something like that here.

PR #25, opened 2017-08-30
Added UserProfile object and validated basic auth and reachability.

adsri:
Added support for UserProfile object and validated basic auth and reachability
while creating the connector.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -332,8 +332,9 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 
 	res = connector
 	// GET UserProfile request is used here to validate connector's basic auth and reachability.
+	var response []UserProfile
 	userprofile := NewUserProfile(UserProfile{})
-	_, err = connector.makeRequest(GET, userprofile, "")
+	err = connector.GetObject(userprofile, "", &response)
 	if err != nil {
 		log.Printf("Failed to connect to the Grid, err: %s \n", err)
 	}
--- a/connector.go
+++ b/connector.go
@@ -331,6 +331,11 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 	connector.Requestor.Init(connector.TransportConfig)
 
 	res = connector
-
+	// GET UserProfile request is used here to validate connector's basic auth and reachability.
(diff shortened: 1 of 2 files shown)


connector.go
```diff
@@ -331,6 +331,11 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 	connector.Requestor.Init(connector.TransportConfig)
 
 	res = connector
-
+	// GET UserProfile request is used here to validate connector's basic auth and reachability.
+	userprofile := NewUserProfile(UserProfile{})
+	_, err = connector.makeRequest(GET, userprofile, "")
```
yuewko: Suggestion: if possible, please use the GetObject() method instead. 

Question:  So, we just need to check that the request is successful? We don't need to check the result?

adsri: @yuewko I am validating two things as below:
1) reachability means verifying if ip and port is reachable 
2) basic authentication means username and password.

For Suggestion: Ok I will do it.

adsri: @yuewko looking at GetObject I feel we dont need to use it in this case, as we donot care about the response of UserProfile.

yuewko: `makeRequest()` is meant to be a service method used by Get/Update/DeleteObject() methods.. For maintainability, we should try to minimize use of service methods.
BTW, I understand that we don't need the result, but the overhead is minimal since we are only doing this once per connection.

I don't have strong opinion on this one. I'll approve it.

adsri: Ok Ko i will do the changes and will consider this from next commits.

connector.go
```diff
@@ -331,6 +331,12 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 	connector.Requestor.Init(connector.TransportConfig)
 
 	res = connector
-
+	// GET UserProfile request is used here to validate connector's basic auth and reachability.
+	var response []UserProfile
+	userprofile := NewUserProfile(UserProfile{})
+	err = connector.GetObject(userprofile, "", &response)
```
adsri: @yuewko Done!

yuewko: Thanks!!

PR #26, opened 2017-09-01
Add WAPI objects

chinmayb:
Fixed the review comments.


chinmayb:
@ishantt  @yuewko My second commit caused some issue, i had pushed akash's changes as well, which got merged later. I created a new pull request.

chinmayb:
Updated the patch with more wapi objects 
@yuewko  Not yet implemented the "feature" you pointed out that day to handle various versions. Will take it up separately.

connector.go
```diff
@@ -338,5 +338,6 @@ func NewConnector(hostConfig HostConfig, transportConfig TransportConfig,
 	if err != nil {
 		log.Printf("Failed to connect to the Grid, err: %s \n", err)
 	}
+
```
ishantt: no needed

connector_test.go
```diff
@@ -4,11 +4,12 @@ import (
...
 	"encoding/json"
 	"fmt"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 	"net/http"
 	"net/url"
 	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
ishantt: any reason for this change?

objects.go
```diff
@@ -63,9 +102,129 @@ type Network struct {
...
+	res := member
+	res.objectType = "member"
+	res.returnFields = returnFields
+	return &res
+}
+
 func NewNetwork(nw Network) *Network {
 	res := nw
 	res.objectType = "network"
+
```
ishantt: not needed

objects.go
```diff
@@ -16,10 +16,15 @@ type EASearch map[string]interface{}
 
 type EADefListValue string
 
+type MultiObjectResult [][]map[string]interface{}
+
+type Args map[string]string
```
ishantt: remove this Args

objects.go
```diff
@@ -16,10 +16,15 @@ type EASearch map[string]interface{}
 
 type EADefListValue string
 
+type MultiObjectResult [][]map[string]interface{}
```
ishantt: Dont use this as a new type . MultiObjectResult can also be a []map[string]interface{}

chinmayb: Done

objects.go
chinmayb: done

connector.go
chinmayb: Done

connector_test.go
chinmayb: Dont think so. Ill remove it

objects.go
```diff
@@ -63,11 +97,130 @@ type Network struct {
...
+	Ref    string `json:"_ref,omitempty"`
+}
+
+type GridResult struct {
+	Ref        string     `json:"_ref,omitempty"`
+	Name       string     `json:"name,omitempty"`
+	NTPSetting NTPSetting `json:"ntp_setting"`
+}
+
+func NewGrid(grid Grid, returnFields []string) *Grid {
```
yuewko: Is it really necessary for `returnFields` to be passed in?

I can see that it might be useful, for efficient reason, to allow user to specify the return fields if the result structure has many large fields that user often don't need. Is that the case?

Still, even if this is the case, the method should set a default set of return fields.

OK... looking at the object manager code, it looks the returnFields are used in the GetXyZ() methods and are always passed the same fixed list of returnFields. I am therefore included to say remove returnFields parameters in this and other NewXyZ() methods and set the defaults inside the methods.

objects.go
```diff
@@ -55,6 +55,40 @@ func NewNetworkView(nv NetworkView) *NetworkView {
...
+	Ipv4Address    string `json:"ipv4_address"`
+	Ipv6Address    string `json:"ipv6_address"`
+	StatusValue    string `json:"status_value"`
+	StepsTotal     int    `json:"steps_total"`
+	StepsCompleted int    `json:"steps_completed"`
+	NodeType       string `json:"type"`
+	Member         string `json:member`
+}
+
+type UpgradeStatusResult struct {
```
yuewko: Is seems that for most of the WAPI objects you are adding (such as this one, UpgradeStatus/UpgradeStatusResult), you have a struct for the object for query, and an object for the query result.  Is this really necessary? What's the reasoning behind that?

objects.go
```diff
@@ -63,11 +97,130 @@ type Network struct {
...
+}
+
+func NewMember(member Member, returnFields []string) *Member {
+	res := member
+	res.objectType = "member"
+	res.returnFields = returnFields
+	return &res
+}
+
 func NewNetwork(nw Network) *Network {
```
yuewko: Consistency...

Originally, the NewXYZ() method is defined right after the definition of the XYZ object.
I have no objection if you like to change it to have all object definitions followed by method definitions.

BUT, please only do it one way.

objects.go
```diff
@@ -63,11 +97,130 @@ type Network struct {
...
+	LanPortSetting map[string]bool        `json:"lan_port_setting,omitempty"`
+}
+
+type NodeInfo struct {
+	HaStatus             string                 `json:"ha_status,omitempty"`
+	HwId                 string                 `json:"hwid,omitempty"`
+	HwModel              string                 `json:"hwmodel,omitempty"`
+	HwPlatform           string                 `json:"hwplatform,omitempty"`
+	HwType               string                 `json:"hwtype,omitempty"`
+	Lan2PhysicalSetting  map[string]bool        `json:"lan2_physical_setting,omitempty"`
```
yuewko: Why is this `map[string]bool`?

According to the WAPI doc, it should be an array of struct like this:

> physicalportsetting : Physical Port Settings.
This structure contains settings for an appliance physical port.
auto_port_setting_enabled
auto_port_setting_enabled

>Enable or disalbe the auto port setting.
Type
Bool.
Create
The default value is undefined.
duplex
duplex

>The port duplex; if speed is 1000, duplex must be FULL.
Type
String.
Valid values are:
FULL
HALF
Create
The default value is undefined.
speed
speed

>The port speed; if speed is 1000, duplex is FULL.
Type
String.
Valid values are:
10
100
1000
Create
The default value is undefined.

objects.go
```diff
@@ -63,11 +97,130 @@ type Network struct {
...
+	LanHaPortSetting     LanHaPortSetting       `json:"lan_Ha_Port_Setting,omitempty"`
+	MgmtNetworkSetting   map[string]interface{} `json:"mgmt_network_setting,omitempty"`
```
yuewko: This has the same issue as above.

Various other fields have the same problem.

object_manager.go
```diff
@@ -387,3 +389,75 @@ func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]
 
 	return result, nil
 }
+
+// GetUpgradeStatus returns the grid upgrade information
+func (objMgr *ObjectManager) GetUpgradeStatus(statusType string) ([]UpgradeStatusResult, error) {
+	res := []UpgradeStatusResult{}
```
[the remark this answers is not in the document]
chinmayb: Done.

object_manager.go
```diff
@@ -387,3 +389,75 @@ func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]
...
+
+	if statusType == "" {
+		// TODO option may vary according to the WAPI version, need to
+		// throw relevant  error.
+		msg := fmt.Sprintf("Status type can not be nil, Valid choices are ['GRID', GROUP', 'VNODE', 'PNODE']")
+		return res, errors.New(msg)
+	}
+	returnFields := []string{"subelements_status,type"}
```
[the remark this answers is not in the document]
chinmayb: yes.

objects.go
chinmayb: because when im creating those objects to make requests
getting below errors
"{ \"Error\": \"AdmConProtoError: Field is not searchable: subelements_status\", ","time":"2017/09/22 15:52:36.801"}

though im not sending those. Not sure why, hence created correpsnding result structs to unmarshal the data.

objects.go
chinmayb: output contains of the below form:

for instance:

"lan_ha_port_setting": {
                    "ha_port_setting": {
                        "auto_port_setting_enabled": true, 
                        "speed": "10"
                    }, 
                    "lan_port_setting": {
                        "auto_port_setting_enabled": true
                    }
                }

so its not array. hence used maps.

objects.go
chinmayb: i dont think its array. here's the sample

"mgmt_physical_setting": {
                    "auto_port_setting_enabled": true
                },

objects.go
chinmayb: i felt  NewXyz should take params, if not, if someone wants specify only subset of returnFields then he will have to modify NewXyz, we better avoid that scenario.

Object_manager can grow with  various getXyz, getXwy, etc.

objects.go
```diff
@@ -63,11 +97,130 @@ type Network struct {
...
+	res.returnFields = returnFields
+	return &res
+}
+
 func NewNetwork(nw Network) *Network {
 	res := nw
 	res.objectType = "network"
-	res.returnFields = []string{"extattrs", "network", "network_view"}
 
+	res.returnFields = []string{"extattrs", "network", "network_view"}
```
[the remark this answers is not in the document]
chinmayb: no need will change it.

objects.go
chinmayb: Done, ill keep it as you mentioned. change it through out to maintain consistency

objects.go
yuewko: It should be a struct, i.e.

`Lan2PhysicalSetting     LanPhysicalSetting`

where,

` type LanPhysicalSetting struct {`
`       AutoPortEnabled       bool`
`       duplex                       string`
etc...

I said "array of struct" because I thought there's multiple of these.

objects.go
yuewko: Same here, should be a struct, i.e.:

`MgmtNetworkSetting     NetworkSettings`

where,

` type NetworkSettings struct {`
`       Address      string   //or however we store ipv4address`
`       Dscp                       uint`
`       Gateway                  string`

etc...

I said "array of struct" because I thought there's multiple of these.

objects.go
yuewko: I don't see why ntp_setting wouldn't work like other fields. Please investigate.

objects.go
yuewko: I agree with you that there may be future use cases that require the ability to modify return fields.

However, that still doesn't justify forcing user to pass in returnFields in NewXyZ(), because returnFields is only relevant if the user wants to do a get/search.

Please consider, instead, adding a method SetReturnFields() for setting returnFields. The NewXyZ() should still define a default list of returnFields that include all the attributes defined for the object.

Note that, realistically, this only allow user to remove fields that he doesn't need returned, since you cannot add any return field that is not already an attribute of the object.

objects.go
chinmayb: Sure, will fix it.

objects.go
chinmayb: Done

objects.go
chinmayb: ok, so you mean to say expand the `IBObject` interface to have SetReturnFields method ? that will enable modifying the _return_fields ?

yuewko: Yes.

chinmayb:
@yuewko i'll update the patch fixing the review comments.

objects.go
chinmayb: other fields are using array of objects and "nil" checks are passing, but in this case it returns an empty struct for which "omitempty" is not working.

yuewko: What is the error?

Do you really need to embed IBBase in the NTPSetting  class?

chinmayb: i dont, i'll remove that, error is the same as i mentioned above. `ntp_setting` is getting queried when doing newgrid, hence it says  "Field is not searchable"

yuewko: OK. You need to add debug to print out the actual WAPI request sent by the connector.

chinmayb:
@yuewko  My bad, updated the patch

objects_test.go
```diff
@@ -9,6 +9,42 @@ import (
 
 var _ = Describe("Objects", func() {
 
+	Context("Grid object", func() {
+
+		grid := Grid{}
+		gridJSON := `{}`
```
yuewko: Testing objects with no data is not sufficient - especially unmarshalling.

objects.go
```diff
@@ -71,6 +103,181 @@ func NewNetwork(nw Network) *Network {
...
+	HaPortSetting  PhysicalPortSetting `json:"ha_port_setting,omitempty"`
+	LanPortSetting PhysicalPortSetting `json:"lan_port_setting,omitempty"`
+	MgmtIpv6addr   string              `json:"mgmt_ipv6addr,omitempty"`
+	MgmtLan        string              `json:"mgmt_lan,omitempty"`
+}
+
+type PhysicalPortSetting struct {
+	AutoPortSettingEnabled bool   `json:"auto_port_setting_enabled"`
+	duplex                 string `json:"duplex,omitempty"`
+	speed                  string
```
ishantt: Shouldnt these be exported fields?

chinmayb: Will do it.

chinmayb:
@adsri @kumaya

objects_test.go
chinmayb: Done

yuewko:
LGTM.

yuewko:
Hold on. @chinmayb you've pushed new changes just now?

chinmayb:
@yuewko  yes, i added couple of more Unit tests cases.

yuewko:
I see.

I don't have any further comments. Please ask your team to review the latest commit. Thanks!

PR #28, opened 2017-09-18
add objects required by Kubernetes external-dns provider

khrisrichardson:
Neither the infobloxopen/infoblox-go-client:master nor the khris.richardson/infoblox-go-client:k8s-external-dns-provider passed for me, but I tried to remain consistent with the other unit tests.

The AllRecords test is commented out, since I suspect it may not pass given that it's a read-only object.

### DIFF
--- a/objects.go
+++ b/objects.go
@@ -138,6 +138,76 @@ func NewUserProfile(userprofile UserProfile) *UserProfile {
 	return &res
 }
 
+type RecordA struct {
+	IBBase   `json:"-"`
+	Ref      string `json:"_ref,omitempty"`
+	Ipv4Addr string `json:"ipv4addr,omitempty"`
+	Name     string `json:"name,omitempty"`
+	View     string `json:"view,omitempty"`
+	Zone     string `json:"zone,omitempty"`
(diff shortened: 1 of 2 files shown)


johnbelamaric:
Thanks! @ishantt @yuewko @adsri please take a look.

Is the `AllRecords` a specific thing only applicable to the K8s integration? Maybe it should move to a `pkg` directory if so.

khrisrichardson:
Hi @johnbelamaric `AllRecords` is a standard Infoblox [object type](https://iib1.dns.lab1.cloudsys.tmcs/wapidoc/objects/allrecords.html). No tomfoolery here.

johnbelamaric:
Ha, ok. Shows what I know :)

khrisrichardson:
Hi @ishantt @yuewko @adsri. When you have a moment to review, I would be extremely grateful. Fortunately there's nothing earth shattering in this PR, as the new objects just follow the lead of the existing objects. I removed the `AllRecords` object as discussed with @johnbelamaric , since I never ended up using it in the Infoblox external DNS provider for Kubernetes.

yuewko:
LGTM.

@johnbelamaric ?

PR #30, opened 2017-10-06
add support for record:host

khrisrichardson:
This PR will facilitate preventing the Infoblox external-dns provider for Kubernetes (WIP) from creating an `A` record if a corresponding `Host` record already exists.


yuewko:
@khrisrichardson - thanks for your contribution!!

PR #50, opened 2018-03-26
Supporting CP-member

AjeyHiremath:


jkraj:
@chinmayb just review this. For now we have add '_proxy_search' in go-client because there is no other way to pass this from caller of this go-client. During our restructure of go-client we can accommodate this things so that caller can decide when to use  '_proxy_search' and when not.

chinmayb:
Sure.

Issue #48, opened 2018-03-14
"Field is not searchable" when getting record with EA


sjberman:
I'm attempting to get a record (either A or CNAME, using GetObject), by specifying an object like this:

```
	obj := ibclient.NewRecordA(
		ibclient.RecordA{
			Name:     "foo.com",
			Ipv4Addr: "1.2.3.4",
			View:     "default",
			Ea:      ibclient.EA{eaKey: eaVal},
		},
	)
```
The body of the HTTP request that gets constructed looks like this;
```
'{"ipv4addr":"1.2.3.4","name":"foo.com","view":"default","extattrs":{"My-EA":{"value":"testing"}}}'
```

However, I'm getting this error: 
```
{ "Error": "AdmConProtoError: Field is not searchable: extattrs",
  "code": "Client.Ibap.Proto",
  "text": "Field is not searchable: extattrs"
}
```
The extattrs are not searchable. I need them to be, as all of the objects I create will have "My-EA", and at a certain point I'll have a general GetObject call to get all objects that I've created, and nothing more. Is this possible?

chinmayb:
You'll have to use 'eaSearch' while creating RecordA object

sjberman:
Can you provide an example? I'm not sure what you mean.

sjberman:
If you're referring to setting the obj.eaSearch to something, the field is not exported, so I don't have access to setting the variable.

chinmayb:
I'll raise a PR soon

sjberman:
Thanks.

I did try to adjust the URL to have the EAKey and EAvalue in it (similar to how you mentioned), but it returned every object with the EA, and seemed to ignore the http body which had the details about the specific object I wanted. In this case I then had to loop through all those objects until I got the one that I cared about.

chinmayb:
PR Raised https://github.com/infobloxopen/infoblox-go-client/pull/49
Can you check whether that works for you ?

sjberman:
In briefly glancing at the code, I see a few things.

1. I need to be able to get both A and CNAME records.
2. The method you added, GetRecordA, accepts an EA and returns all A records. Which means I'll have to parse through and find the one I want (unless I want all of them, which I do in some cases). One of the nice things about the current GetObject method is that I can specify an A or CNAME record with specific fields, and I'll get back just that record. Being able to simply add an EA to the object I've constructed, and calling GetObject on that should return just that object (or a list of just that object, as it currently works).

[...]

PR #51, opened 2018-04-02
Moving the EA filter query to request URL instead of building it on t…

chinmayb:
…he payload body

    * Currently the EA filters are passed as api payload, this is creating
      problem when the obj is empty and ea is requested in GET call. Hence
      moving it as URL parameter.

      `objJSON = append(append(objJSON[:len(objJSON)-1], byte(',')), eaSearchJSON[1:]...)`

    * Add Get A record call

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ _testmain.go
 *.exe
 *.test
 *.prof
+*.coverage
--- a/connector.go
+++ b/connector.go
@@ -61,7 +61,7 @@ func NewTransportConfig(sslVerify string, httpRequestTimeout int, httpPoolConnec
 
 type HttpRequestBuilder interface {
 	Init(HostConfig)
-	BuildUrl(r RequestType, objType string, ref string, returnFields []string) (urlStr string)
+	BuildURL(r RequestType, obj IBObject, ref string) (urlStr string)
 	BuildBody(r RequestType, obj IBObject) (jsonStr []byte)
 	BuildRequest(r RequestType, obj IBObject, ref string) (req *http.Request, err error)
 }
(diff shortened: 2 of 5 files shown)


yuewko:
I assume this PR is to address https://github.com/infobloxopen/infoblox-go-client/issues/48. If so, please indicate accordingly in the description.

Please refer to https://github.com/infobloxopen/infoblox-go-client/issues/48#issuecomment-378324915 - I don't believe this is the recommended implementation described in the issue.

chinmayb:
> If  the user did not set any EA in the object, then the search would not include EA
Yes.  Currently if only EA is set and user calls "Getobject" method EAs are not included for searching.  Because `EASearch` field needs to be filled. EA field is specific to an object unlike easearch which is a part of IBBase.

yuewko:
Removing EaSearch would not lead to client changes, since EaSearch is only used internally.

Now, for EA, the code need to be able to marshal it differently depending on whether it is used for search or set/update. If this can be achieved leaving it the way it is (i.e. just a map), that would be perfect as it wouldn't require client changes. Maybe give this more thoughts?

chinmayb:
> Removing  EaSearch would not lead to client changes
Yes, I was referring  to  changing the EA to struct type.

yuewko:
For backward compatibility, it would be good to leave EA just a map as it is now.

One idea is to add a "private" entry to the map to indicate whether it is used for update or search, and later on marshal it accordingly.

Of course, please consider other options. If changing to struct make it more future proof, don't write it off solely for backward compatibility reason.

chinmayb:
A you know ea is defined as the following for an obj
`Ea     EA     `json:"extattrs,omitempty"``

so if Ea field also needs to be used for searching , it should be set empty. If not `extattrs` will be marshaled and server will error out saying "Field is not searchable: extattrs".

it'll involve more hacks for checking the method type in build body/Get object

yuewko:
@chinmayb , I took a deep dive into this today....  I came up with a couple of workable solutions, but none of them is perfect and all require some (albeit minor) external API changes. If you are interested, I'll be more than happy to discuss this with you offline.

Also, I'm pretty sure that the `GetRecordA()` method implemented in this PR is not what @sjberman is looking for. According to his comment https://github.com/infobloxopen/infoblox-go-client/issues/48#issuecomment-374287770, he would like to specify EA **(along with other fields**). 

Unless we (meaning you ;-) ) come up with a much improved approach in the meantime, I propose closing this PR and take the time to **do it right** - since I suspect some overhaul and external API change might be inevitable.

I apologize for the prolonged back and forth.

chinmayb:
@yuewko Yes, the PR doesn't exactly does what @sjberman wants. It provides an alternative way.  I suppose we can close this. We'll discuss this offline.

PR #54, opened 2018-04-11
Updation of name field in FixedAddress record

jkraj:
Presently 'name' field is left out so updating that field with
vmID value


yuewko:
Just curious, is this a requested feature?

The VM ID is not very descriptive, so, IMHO, it's not particular useful for presentation purposes. It is already available as EA for search purposes.

PR #57, opened 2018-04-26
Added GetNetworks method and TENANT-ID key 

AjeyHiremath:
Added GetNetworks method to get all the networks at object_manager.go and added TENANT-ID key  at lock.go

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -24,6 +24,7 @@ type IBObjectManager interface {
 	GetEADefinition(name string) (*EADefinition, error)
 	CreateEADefinition(eadef EADefinition) (*EADefinition, error)
 	UpdateNetworkViewEA(ref string, addEA EA, removeEA EA) error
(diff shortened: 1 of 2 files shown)


object_manager.go
```diff
@@ -536,3 +537,26 @@ func (objMgr *ObjectManager) GetGridInfo() ([]Grid, error) {
 	err := objMgr.connector.GetObject(gridObj, "", &res)
 	return res, err
 }
+
+// Get the all networks
+func (objMgr *ObjectManager) GetNetworks(netview string, cidr string, ea EA) (*[]Network, error) {
```
yuewko: There is already a `GetNetwork()` method which performs essentially the same function, except that it returns only the first network. So, I suggest doing one of two things:
- Have only one `GetNetwork()` method that returns multiple networks
- If the reasoning is to prevent breaking API, then at least have `GetNetwork()` uses `GetNetworks()`, instead of duplicating the code.

PR #58, opened 2018-05-10
apply request timeout to http.Client

r0fls:
Right now the `cfg.HttpRequestTimeout` is only being applied to the `http.Transport.ResponseHeaderTimeout`. However, it should be applied to the entire `http.Client`. See the Client Timeout section of https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ for more information.

`http.Client` reference: https://golang.org/pkg/net/http/#Client

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -139,7 +139,7 @@ func (whr *WapiHttpRequestor) Init(cfg TransportConfig) {
 		log.Fatal(err)
 	}
 
-	whr.client = http.Client{Jar: jar, Transport: tr}
+	whr.client = http.Client{Jar: jar, Transport: tr, Timeout: cfg.HttpRequestTimeout * time.Second}
 }
 
 func (whr *WapiHttpRequestor) SendRequest(req *http.Request) (res []byte, err error) {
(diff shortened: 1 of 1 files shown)


yuewko:
@r0fls , thanks for submitting the PR! I believe it makes lots of sense to set the `http.Client Timeout`.

Having said that, I don't think it make sense to use the same timeout for `http.Transport ResponseHeaderTimeout` as `http.Client Timeout`. i.e. we should either have separate configs for those, or, stick with just `http.Client Timeout`.

@jkraj , could you please chime in on this?

r0fls:
Thanks for the review @yuewko! I can remove the transport timeout setting if preferred. Given the variable is called `cfg.HttpRequestTimeout`, I think it's misleading to only set the `ResponseHeaderTimeout`.

yuewko:
Yes, @r0fls,  I prefer remove  the `http.Transport RequestHeaderTimeout`. Thanks!!

r0fls:
Done!

PR #63, opened 2018-05-24
forced Proxy test case

AjeyHiremath:
Added forced proxy implementation in connector_test.go

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -92,7 +92,7 @@ var _ = Describe("Connector", func() {
 				objType := "networkview"
 				ref := ""
 				returnFields := []string{}
-				forcedProxy := false
+				var forcedProxy bool
 				expectedURLStr := fmt.Sprintf("https://%s:%s/wapi/v%s/%s",
 					host, port, version, objType)
 				urlStr := wrb.BuildUrl(CREATE, objType, ref, returnFields, forcedProxy)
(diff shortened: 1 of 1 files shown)


AjeyHiremath:
@yuewko Now Covers both the cases where forcedProxy == false and forcedProxy == true where request type is GET, In other places it will be false only.

connector_test.go
```diff
@@ -102,11 +102,70 @@ var _ = Describe("Connector", func() {
...
+					res:    []byte(fakeref),
+				}
+				OrigValidateConnector := ValidateConnector
+				ValidateConnector = MockValidateConnector
+				defer func() { ValidateConnector = OrigValidateConnector }()
+				conn, err := NewConnector(hostConfig, transportConfig,
+					frb, fhr)
+				if err != nil {
+					Fail("Error creating Connector")
+				}
```
yuewko: This block is for testing "BuildUrl" - see Line 90.

Line 152-159 is really a just test for Connector instantiation. This should be done in a separate block of tests.

connector_test.go
```diff
@@ -102,11 +102,70 @@ var _ = Describe("Connector", func() {
...
+				var forcedProxy bool
 				expectedURLStr := fmt.Sprintf("https://%s:%s/wapi/v%s/%s?%s",
 					host, port, version, objType, returnFieldsStr)
+				_, err = conn.Requestor.SendRequest(httpReq)
```
yuewko: Again, this block of test in for "BuildUrl". This doesn't belong here.

And, btw, the only thing the fake `SendRequest()` method does is to check that the argument is the same - not sure it is very useful.

connector_test.go
```diff
@@ -102,11 +102,70 @@ var _ = Describe("Connector", func() {
...
+				if err != nil {
+					forcedProxy = true
+					expectedURLStr = fmt.Sprintf("https://%s:%s/wapi/v%s/%s?%s&%s",
+						host, port, version, objType, qry, returnFieldsStr)
+				}
```
yuewko: OK, as it is, `err` is ALWAYS `nil`. So, this block is NEVER executed.

So, we are still not testing the case for `forcedProxy == true`.

connector_test.go
```diff
@@ -115,7 +174,7 @@ var _ = Describe("Connector", func() {
 				objType := ""
 				ref := "fixedaddress/ZG5zLmJpbmRfY25h:12.0.10.1/external"
 				returnFields := []string{}
-				forcedProxy := false
+				var forcedProxy bool
```
yuewko: Why this change? The original `forcedProxy := false` is much clearer.

(There are other instances of this change too).

PR #67, opened 2018-05-25
forced proxy test cases

AjeyHiremath:
added forced proxy test cases for both the values true and false


somashekhar:
Closing the PR since the respective changes are already in place.

Issue #62, opened 2018-05-24
FixedAddress: "Name" field is not filled in when searching


eest:
Hello,

Starting with version 0.5.0 the FixedAddress struct contains a "Name" field which I had use for. I noticed however that the Name field will always be empty when searching. This is because at least in my WAPI version the wapidoc states that only "ipv4addr" and  "network_view" is included in the basic object and `NewFixedAddress()` does not include "name" when filling in the `returnFields` slice:
```
func NewFixedAddress(fixedAddr FixedAddress) *FixedAddress {
        res := fixedAddr
        res.objectType = "fixedaddress"
        res.returnFields = []string{"extattrs", "ipv4addr", "mac", "network", "network_view"}

        return &res
}
```

Since `returnFields` is unexported I can not add "name" from my own code. I noticed there seemed to exist some basic idea of having a function to modify this slice because of a commented out `SetReturnFields()` function:
```
type IBObject interface {
        ObjectType() string
        ReturnFields() []string
        EaSearch() EASearch
        //SetReturnFields([]string)
}
```

... but I see no further traces of such code.

eest:
@jkraj: I guess you should have more input on this since you added the "Name" field in the struct.

jkraj:
@eest Since **name** is not searchable I didn't add that into _returnFields_ slice. If I include it will solve your issue.

chinmayb:
https://github.com/infobloxopen/infoblox-go-client/pull/69

Issue #59, opened 2018-05-16
Recent update removed "Ipv4Addr" field from RecordHost (now HostRecord)


eest:
Hello,

I tried updating to 0.6.0 and it broke my code. The initial problem was that Host record related structs had been renamed from "RecordHost[...]" to "HostRecord[...]". While I guess that "HostRecord" reads a bit more naturally it's a bit strange since all other record types are still called "RecordA", "RecordCNAME" etc.

I can live with that change, but I also noticed the now renamed HostRecord struct no longer contains the field "Ipv4Addr". I have been using that to search for hosts by IP address like so:
```
var hostResult []ibclient.RecordHost

recordHostObj := ibclient.NewRecordHost(ibclient.RecordHost{
    Ipv4Addr: ip,
})

err := conn.GetObject(recordHostObj, "", &hostResult)
```

This of course no longer works. Was this removed on purpose and I am supposed to fill in an Ipv4Addrs slice instead or should it be added back?

Pinging @saiprasannasastry and @jkraj since you were involved in changing the code.

jkraj:
@eest Yes it is modified intentionally.  Present structure represents actual wapi object representation. So you should use Ipv4Addrs slice.

chinmayb:
Looks like we need to keep "Ipv4addr" field along with "Ipv4addrs"  as well @jkraj .  If you see the doc "wapidoc/objects/record.host.html#ipv4addr" Ipv4addr of string type is searchable, while ipv4addrs is not.

eest:
@chinmayb: I applied the diff to my 0.6.0 branch and it made my search work, thanks!

eest:
I noticed #69 has been merged, it would be great if you could create a new tag when this is also solved so i have a version that supports both host searching by ip address and the FixedAddress name addition from 0.5.0 :).

chinmayb:
@eest you can use 0.7.0 version.

eest:
@chinmayb: I can not, because 0.7.0 does not include the fix for FixedAddress added in #69.

chinmayb:
I had updated the tag by force pushing. It didnt seem to work. I created a new release v0.7.1. Please use that

eest:
@chinmayb the 0.7.1 tag seems to work as expected, thanks!

PR #69, opened 2018-05-25
'name' field added in Fixedaddress struct

AjeyHiremath:
added field ‘name’ in return_field slice of Fixedaddress struct

### DIFF
--- a/objects.go
+++ b/objects.go
@@ -306,7 +306,7 @@ type FixedAddress struct {
 func NewFixedAddress(fixedAddr FixedAddress) *FixedAddress {
 	res := fixedAddr
 	res.objectType = "fixedaddress"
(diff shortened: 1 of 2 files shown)


chinmayb:
Can you also add "ipv4addr" to host record
https://github.com/infobloxopen/infoblox-go-client/issues/59.
Since both are regressions, we can fix in this commit only may be.

yuewko:
https://github.com/infobloxopen/infoblox-go-client/pull/69#issuecomment-392254690
```
Can you also add "ipv4addr" to host record
#59.
Since both are regressions, we can fix in this commit.
```
It would be better to have a separate commit for #59, since #62 that this PR fixes is very much independent.

yuewko:
Please make sure CI passes.

PR #71, opened 2018-06-01
CP member to get  GM objects

AjeyHiremath:
CP member using  the objects(ex: network view) created in Grid Master

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -247,9 +247,8 @@ func (wrb *WapiRequestBuilder) BuildRequest(t RequestType, obj IBObject, ref str
 	return
 }
 
-func (c *Connector) makeRequest(t RequestType, obj IBObject, ref string) (res []byte, err error) {
+func (c *Connector) makeRequest(t RequestType, obj IBObject, ref string, forcedProxy bool) (res []byte, err error) {
 	var req *http.Request
-	var forcedProxy bool
 	req, err = c.RequestBuilder.BuildRequest(t, obj, ref, forcedProxy)
 	res, err = c.Requestor.SendRequest(req)
 	if err != nil {
@@ -264,8 +263,8 @@ func (c *Connector) makeRequest(t RequestType, obj IBObject, ref string) (res []
 
 func (c *Connector) CreateObject(obj IBObject) (ref string, err error) {
 	ref = ""
-
-	resp, err := c.makeRequest(CREATE, obj, "")
+	var forcedProxy bool
+	resp, err := c.makeRequest(CREATE, obj, "", forcedProxy)
 	if err != nil || len(resp) == 0 {
 		log.Printf("CreateObject request error: '%s'\n", err)
 		return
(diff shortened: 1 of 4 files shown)


AjeyHiremath:
@yuewko @chinmayb @adsri

connector.go
```diff
@@ -264,8 +263,8 @@ func (c *Connector) makeRequest(t RequestType, obj IBObject, ref string) (res []
 
 func (c *Connector) CreateObject(obj IBObject) (ref string, err error) {
 	ref = ""
-
-	resp, err := c.makeRequest(CREATE, obj, "")
+	var forcedProxy bool
```
yuewko: It would be much clearer doing `forcedProxy := false` instead.

connector.go
```diff
@@ -281,29 +280,37 @@ func (c *Connector) CreateObject(obj IBObject) (ref string, err error) {
 }
 
 func (c *Connector) GetObject(obj IBObject, ref string, res interface{}) (err error) {
-	resp, err := c.makeRequest(GET, obj, ref)
+	var forcedProxy bool
```
yuewko: ditto

connector.go
```diff
@@ -281,29 +280,37 @@ func (c *Connector) CreateObject(obj IBObject) (ref string, err error) {
...
 	}
-
 	return
 }
 
 func (c *Connector) DeleteObject(ref string) (refRes string, err error) {
 	refRes = ""
-
-	resp, err := c.makeRequest(DELETE, nil, ref)
+	var forcedProxy bool
```
yuewko: ditto

connector.go
```diff
@@ -319,9 +326,9 @@ func (c *Connector) DeleteObject(ref string) (refRes string, err error) {
 }
 
 func (c *Connector) UpdateObject(obj IBObject, ref string) (refRes string, err error) {
-
+	var forcedProxy bool
```
yuewko: ditto

connector.go
```diff
@@ -339,7 +346,8 @@ func (c *Connector) UpdateObject(obj IBObject, ref string) (refRes string, err e
 // be used in a defer statement after the Connector has been successfully
 // initialized.
 func (c *Connector) Logout() (err error) {
-	_, err = c.makeRequest(CREATE, nil, "logout")
+	var forcedProxy bool
```
yuewko: ditto

object_manager.go
```diff
@@ -458,8 +458,8 @@ func (objMgr *ObjectManager) DeleteHostRecord(ref string) (string, error) {
 func (objMgr *ObjectManager) CreateMultiObject(req *MultiRequest) ([]map[string]interface{}, error) {
 
 	conn := objMgr.connector.(*Connector)
-
-	res, err := conn.makeRequest(CREATE, req, "")
+	var forcedProxy bool
```
yuewko: ditto

connector.go
```diff
@@ -281,29 +280,37 @@ func (c *Connector) CreateObject(obj IBObject) (ref string, err error) {
 }
 
 func (c *Connector) GetObject(obj IBObject, ref string, res interface{}) (err error) {
```
chinmayb: since we  are changing the method signatures, why not add options instead of just forcedProxy. 
Since we may need to include other query params , like "max_results/paging" etc.

may be like 
take another param, ...option in GetObject
`
type Options struct {
       forceProxy bool
       returnFields []string
       eaForSearch  bool
       max_results int
       paging bool
}

type option func(*Options)
`
we can build query params using the options while doing get. What say @yuewko @jkraj

AjeyHiremath:
@sganesh-infoblox

connector.go
AjeyHiremath: done

connector.go
AjeyHiremath: done

object_manager.go
AjeyHiremath: done

AjeyHiremath:
@adsri 
@sganesh-infoblox

connector_test.go
```diff
@@ -92,34 +92,53 @@ var _ = Describe("Connector", func() {
...
 				expectedURLStr := fmt.Sprintf("https://%s:%s/wapi/v%s/%s",
 					host, port, version, objType)
-				urlStr := wrb.BuildUrl(CREATE, objType, ref, returnFields, forcedProxy)
+			FORCED_PROXY:
+				urlStr := wrb.BuildUrl(CREATE, objType, ref, returnFields, queryParams)
 				Expect(urlStr).To(Equal(expectedURLStr))
+				if queryParams.forceProxy == false {
+					queryParams.forceProxy = true // proxy enabled
+					goto FORCED_PROXY
+				}
```
yuewko: I'm afraid I don't see any justifiable reason for using goto here (L98-L104):
- if the test did fail for any reason, you have no way of telling which case (forceProxy true/false) caused the failure.
- there should be two different test cases, one for `forceProxy := true` and one for `forceProxy:=false`, and the two cases should be inside different instances of `It()` blocks.
- even the implementation approach in L108-L126 would have been much more preferable.

AjeyHiremath: I have addressed all the comments in PR "modified test cases for force proxy  #72"

PR #72, opened 2018-06-05
modified test cases for force proxy

AjeyHiremath:
There are two different test cases, one for forceProxy := true and one for forceProxy:=false, and the two cases are  inside different instances of It() blocks.

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -88,75 +88,70 @@ var _ = Describe("Connector", func() {
 		wrb := WapiRequestBuilder{HostConfig: hostCfg}
 
 		Describe("BuildUrl", func() {
-			It("should return expected url string for CREATE request", func() {
+			Context("for CREATE request", func() {
 				objType := "networkview"
 				ref := ""
 				returnFields := []string{}
-				queryParams := QueryParams{forceProxy: false} //disable proxy
-				expectedURLStr := fmt.Sprintf("https://%s:%s/wapi/v%s/%s",
-					host, port, version, objType)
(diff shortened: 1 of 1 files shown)


connector_test.go
```diff
@@ -92,16 +92,21 @@ var _ = Describe("Connector", func() {
...
 				ref := ""
 				returnFields := []string{}
-				queryParams := QueryParams{forceProxy: false}
+				queryParams := QueryParams{forceProxy: false} //disable proxy
+				expectedURLStr := fmt.Sprintf("https://%s:%s/wapi/v%s/%s",
+					host, port, version, objType)
+				urlStr := wrb.BuildUrl(CREATE, objType, ref, returnFields, queryParams)
+				Expect(urlStr).To(Equal(expectedURLStr))
+			})
+			It("should return expected url string for CREATE request", func() {
```
yuewko: The description in L91 and L101 should state EXACTLY what it's testing, i.e. L91 should say `should return expected url string for CREATE request when forceProxy is false` and L101 should say `should return expected url string for CREATE request when forceProxy is true`.

The way it is now, again, if the test fails you CANNOT tell from the output which test failed.

Ideally, the tests should be structured this way:

```
Describe("BuildUrl", func() {
    :
    :
    Context("for CREATE request", func() {
        :
        :
        It("should return expected url string when forceProxy is true", func() {
           :
           :
        })
        :
        :
        It("should return expected url string when forceProxy is false", func() {
           :
           :
        })
    })
})
```

AjeyHiremath: done

PR #74, opened 2019-01-14
DNS Support for go-client

saiprasannasastry:
Added support for CNAME,A and PTR Records.
Enabled to create Host Recordwith DNS.
Added UT's

UT Results

Running Suite: InfobloxGoClient Suite
=====================================
Random Seed: 1546495859
Will run 130 of 130 specs

••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••
••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••
Ran 130 of 130 Specs in 0.003 seconds
SUCCESS! -- 130 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
ok      github.com/infobloxopen/infoblox-go-client      0.024s

@jkraj please review


object_manager.go
```diff
@@ -376,12 +410,14 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
 	return newEadef, err
 }
 
-func (objMgr *ObjectManager) CreateHostRecordWithoutDNS(recordName string, netview string, cidr string, ipAddr string, macAddress string, vmID string) (*HostRecord, error) {
+func (objMgr *ObjectManager) CreateHostRecord(enabledns bool, recordName string, netview string, dnsview string, cidr string, ipAddr string, macAddress string, vmID string, vmName string) (*HostRecord, error) {
```
chinmayb: You can split this into multiple functions rather than just having `CreateHostRecord`. This will have more if-else conditions as we find more cases. You can do this

Make `createHostRecord` generic which takes `extrattrs` etc.

Then have 
`CreateHostRecordWithoutDNS`
`CreateHostRecordForIp`
`CreateHostRecordFromRange` and so on which will call `createHostRecord`

object_manager.go
```diff
@@ -413,7 +450,7 @@ func (objMgr *ObjectManager) GetHostRecord(ref string) (*HostRecord, error) {
 	return recordHost, err
 }
 
-func (objMgr *ObjectManager) GetHostRecordWithoutDNS(recordName string, netview string, cidr string, ipAddr string) (*HostRecord, error) {
+func (objMgr *ObjectManager) GetHostRecordWithoutref(recordName string, netview string, cidr string, ipAddr string) (*HostRecord, error) {
```
chinmayb: i feel `GetHostRecordByName` would  be better. May be change the above to `GetHostRecordWithRef`

object_manager.go
```diff
@@ -435,7 +472,7 @@ func (objMgr *ObjectManager) GetIpAddressFromHostRecord(host HostRecord) (string
 	return host.Ipv4Addrs[0].Ipv4Addr, err
 }
 
-func (objMgr *ObjectManager) UpdateHostRecordWithoutDNS(hostRref string, ipAddr string, macAddress string, vmID string) (string, error) {
+func (objMgr *ObjectManager) UpdateHostRecord(hostRref string, ipAddr string, macAddress string, vmID string) (string, error) {
```
chinmayb: dont you need `vmName` just like above here?

object_manager.go
```diff
@@ -454,6 +491,102 @@ func (objMgr *ObjectManager) DeleteHostRecord(ref string) (string, error) {
 	return objMgr.connector.DeleteObject(ref)
 }
 
+func (objMgr *ObjectManager) CreateARecord(netview string, dnsview string, recordname string, cidr string, ipAddr string, vmID string, vmName string) (*RecordA, error) {
```
chinmayb: accept `EA` as a parameter, populate if its empty.

object_manager.go
```diff
@@ -317,20 +336,35 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
+func (objMgr *ObjectManager) UpdateFixedAddress(fixedAddrRef string, matchclient string, macAddress string, vmID string, vmName string) (*FixedAddress, error) {
 
 	updateFixedAddr := NewFixedAddress(FixedAddress{Ref: fixedAddrRef})
 
 	if len(macAddress) != 0 {
 		updateFixedAddr.Mac = macAddress
 	}
 
-	if vmID != "" {
+	if vmID != "" &&vmName != ""{
```
chinmayb: This could be `||`, is it not ? or are you expecting both not empty ?

object_manager.go
```diff
@@ -317,20 +336,35 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
+        fixedAddr := NewFixedAddress(FixedAddress{})
+        err := objMgr.connector.GetObject(fixedAddr, ref, &fixedAddr)
+        return fixedAddr, err
+}
+
+func (objMgr *ObjectManager) DeleteFixedAddress(ref string) (string, error) {
+        return objMgr.connector.DeleteObject(ref)
+}
+
+func (objMgr *ObjectManager) UpdateFixedAddress(fixedAddrRef string, matchclient string, macAddress string, vmID string, vmName string) (*FixedAddress, error) {
```
chinmayb: what does `matchclient` do ?

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
...
+	GetFixedAddressFromref(ref string) (*FixedAddress, error)
+	DeleteFixedAddress(ref string) (string, error)
 	ReleaseIP(netview string, cidr string, ipAddr string, macAddr string) (string, error)
 	DeleteNetwork(ref string, netview string) (string, error)
 	GetEADefinition(name string) (*EADefinition, error)
 	CreateEADefinition(eadef EADefinition) (*EADefinition, error)
 	UpdateNetworkViewEA(ref string, addEA EA, removeEA EA) error
+	CreateHostRecord(enabledns bool, recordName string, netview string, dnsview string, cidr string, ipAddr string, macAddress string, vmID string, vmName string) (*HostRecord, error)
+	GetHostRecord(ref string) (*HostRecord, error)
+	GetHostRecordWithoutref(recordName string, netview string, cidr string, ipAddr string) (*HostRecord, error)
```
chinmayb: This is getting huge, you coud rather group ARecord/hostrecord into a separate interface and add it here.

object_manager.go
saiprasannasastry: so , when we create a Reservation by giving 00:00:00:00:00:00 using create fixed address call and update this reserved address to fixed address then the match client flag should be enabled.

object_manager.go
saiprasannasastry: actually the enableDns flag here decides if we need to create a host Record for DNS purposes or not.
It is like this , if we pass true for enable dns then we will have a host record created under that particular view and if we dont then it will just create a hostrecord for IPAM purposes

object_manager.go
saiprasannasastry: actually i am expecting both to be not empty , because even if one of them is passed empty , it raises wapi error.
Tested when i made it ||
case 1 : passed vmID but no name
 "Error": "AdmConDataError: None (IBDataConflictError: IB.Data.Conflict:Missing value for extensible attribute 'VM Name'.)",

case 2 :passed vmName but no ID
 "Error": "AdmConDataError: None (IBDataConflictError: IB.Data.Conflict:Missing value for extensible attribute 'VM ID'.)",

object_manager.go
saiprasannasastry: yes , i have missed vmName, thanks for pointing out

object_manager.go
saiprasannasastry: okay ,will do it

connector.go
```diff
@@ -283,12 +283,14 @@ func (c *Connector) GetObject(obj IBObject, ref string, res interface{}) (err er
...
-	var result []map[string]interface{}
+	var result interface{}
 	err = json.Unmarshal(resp, &result)
 	if err != nil {
 		log.Printf("Cannot unmarshall to check empty value '%s', err: '%s'\n", string(resp), err)
 	}
-	if resp == nil || len(result) == 0 {
+
+        var data []interface{}
+	if resp == nil || (reflect.TypeOf(result) == reflect.TypeOf(data) && len(result.([]interface{})) == 0) {
```
yuewko: It's not clear what the reason is for the changes in GetObject(). Could you please elaborate?

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
 	GetNetworkView(name string) (*NetworkView, error)
 	GetNetwork(netview string, cidr string, ea EA) (*Network, error)
 	GetNetworkContainer(netview string, cidr string) (*NetworkContainer, error)
-	AllocateIP(netview string, cidr string, ipAddr string, macAddress string, name string, vmID string) (*FixedAddress, error)
+	AllocateIP(netview string, cidr string, ipAddr string, macAddress string, name string, vmID string, vmName string) (*FixedAddress, error)
 	AllocateNetwork(netview string, cidr string, prefixLen uint, name string) (network *Network, err error)
-	UpdateFixedAddress(fixedAddrRef string, macAddress string, name string, vmID string) (*FixedAddress, error)
+	UpdateFixedAddress(fixedAddrRef string, matchclient string, macAddress string, vmID string, vmName string) (*FixedAddress, error)
 	GetFixedAddress(netview string, cidr string, ipAddr string, macAddr string) (*FixedAddress, error)
+	GetFixedAddressFromref(ref string) (*FixedAddress, error)
```
yuewko: Typically, it would have been called `GetFixedAddressByRef`.

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
...
+	DeleteFixedAddress(ref string) (string, error)
 	ReleaseIP(netview string, cidr string, ipAddr string, macAddr string) (string, error)
 	DeleteNetwork(ref string, netview string) (string, error)
 	GetEADefinition(name string) (*EADefinition, error)
 	CreateEADefinition(eadef EADefinition) (*EADefinition, error)
 	UpdateNetworkViewEA(ref string, addEA EA, removeEA EA) error
+	CreateHostRecord(enabledns bool, recordName string, netview string, dnsview string, cidr string, ipAddr string, macAddress string, vmID string, vmName string) (*HostRecord, error)
+	GetHostRecord(ref string) (*HostRecord, error)
```
yuewko: The naming of the method should be consistent. Example for Fixed Address:

- `GetFixedAddress(attr1, attr2)`
- `GetFixedAddressByRef(ref)`

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
...
+	GetHostRecordWithoutref(recordName string, netview string, cidr string, ipAddr string) (*HostRecord, error)
+	GetIpAddressFromHostRecord(host HostRecord) (string, error)
+	UpdateHostRecord(hostRref string, ipAddr string, macAddress string, vmID string) (string, error)
+	DeleteHostRecord(ref string) (string, error)
+	CreateARecord(netview string, dnsview string, recordname string, cidr string, ipAddr string, vmID string, vmName string) (*RecordA, error)
+        GetARecord(ref string) (*RecordA, error)
```
yuewko: (ditto)

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
...
+	DeleteARecord(ref string) (string, error)
+	CreateCNAMERecord(canonical string, recordname string ,dnsview string,)(*RecordCNAME,error)
+        GetCNAMERecord(ref string) (*RecordA, error)
```
yuewko: (ditto)

object_manager.go
```diff
@@ -15,15 +15,32 @@ type IBObjectManager interface {
...
+	DeleteCNAMERecord(ref string) (string, error)
+	CreatePTRRecord(netview string, dnsview string, recordname string, cidr string, ipAddr string, vmID string, vmName string) (*RecordPTR, error)
+	GetPTRRecord(ref string) (*RecordPTR, error)
```
yuewko: (ditto)

object_manager.go
```diff
@@ -246,15 +263,17 @@ func GetIPAddressFromRef(ref string) string {
...
+func (objMgr *ObjectManager) AllocateIP(netview string, cidr string, ipAddr string, macAddress string, name string, vmID string, vmName string) (*FixedAddress, error) {
 	if len(macAddress) == 0 {
 		macAddress = MACADDR_ZERO
 	}
 
 	ea := objMgr.getBasicEA(true)
 	ea["VM ID"] = "N/A"
-	if vmID != "" {
+	ea["VM Name"] = "N/A"
+	if vmID != "" && vmName != "" {
```
yuewko: Why only when BOTH are not `""`?

object_manager.go
```diff
@@ -454,6 +491,102 @@ func (objMgr *ObjectManager) DeleteHostRecord(ref string) (string, error) {
...
+func (objMgr *ObjectManager) CreateCNAMERecord(canonical string, recordname string ,dnsview string,)(*RecordCNAME,error) {
+
+	recordCNAME := NewRecordCNAME(RecordCNAME{
+		View:      dnsview,
+		Name:      recordname,
+		Canonical: canonical})
+
+	ref, err := objMgr.connector.CreateObject(recordCNAME)
+	recordCNAME.Ref = ref
+	err = objMgr.connector.GetObject(recordCNAME, ref, &recordCNAME)
```
yuewko: (similar comment to above)

object_manager.go
```diff
@@ -454,6 +491,102 @@ func (objMgr *ObjectManager) DeleteHostRecord(ref string) (string, error) {
...
+		Ea:          ea})
+
+	if ipAddr == "" {
+		recordA.Ipv4Addr = fmt.Sprintf("func:nextavailableip:%s,%s", cidr, netview)
+	} else {
+		recordA.Ipv4Addr = ipAddr
+	}
+	ref, err := objMgr.connector.CreateObject(recordA)
+	recordA.Ref = ref
+	err = objMgr.connector.GetObject(recordA, ref, &recordA)
```
yuewko: Wouldn't `GetObject()` be ONLY necessary if `ipAddr` passed in is `""`? 

IOW, in the case if `ipAddr` is specified, wouldn't the resulting `recordA` would be no different from the `recordA` before calling `GetObject()`?

object_manager.go
```diff
@@ -454,6 +491,102 @@ func (objMgr *ObjectManager) DeleteHostRecord(ref string) (string, error) {
...
+		Ea:          ea})
+
+	if ipAddr == "" {
+		recordPTR.Ipv4Addr = fmt.Sprintf("func:nextavailableip:%s,%s", cidr, netview)
+	} else {
+		recordPTR.Ipv4Addr = ipAddr
+	}
+	ref, err := objMgr.connector.CreateObject(recordPTR)
+	recordPTR.Ref = ref
+	err = objMgr.connector.GetObject(recordPTR, ref, &recordPTR)
```
yuewko: (ditto)

objects.go
```diff
@@ -300,6 +300,7 @@ type FixedAddress struct {
 	IPAddress   string `json:"ipv4addr,omitempty"`
 	Mac         string `json:"mac,omitempty"`
 	Name        string `json:"name,omitempty"`
+	MatchClient string `json:"match_client,omitempty"`
```
yuewko: Shouldn't there be a corresponding change in unit tests as a result of adding this field?

object_manager.go
chinmayb: ok then it should be like this
```suggestion
	if vmID != "" {
	   ea["VM ID"] = vmID
	}
	if vmName != ""{
	   ea["VM Name"] = vmName
	}
```

connector.go
saiprasannasastry: Because of forced proxy implementation , it was assumed that the result/response we get was either nil or list. but when a response which is of type dictionary was coming it was failing. this was done to  handle the case when a dictionary was coming wasn't implemented.

object_manager.go
saiprasannasastry: will do

object_manager.go
saiprasannasastry: will do

object_manager.go
saiprasannasastry: will do

object_manager.go
saiprasannasastry: will do

objects.go
saiprasannasastry: just like name , even match client can be empty and not necessary to be passed.Hence did not add

object_manager.go
saiprasannasastry: yes, makes sense .Let me test it out without GetObject()

objects.go
yuewko: So, we should at least test the case when it IS passed in.

object_manager.go
yuewko: ok

connector.go
yuewko:
```
it was assumed that the result/response we get was either nil or list
```
I'm afraid I don't understand. In the original code, result IS a map: `var result []map[string]interface{}`

BTW, can you also include example responses that failed in the original code please?

objects.go
saiprasannasastry: okay, will add it

connector.go
yuewko: I'm afraid it is still not clear.

IIRC, the format of the response is no different whether the request is proxied or not. I.e. the response is always something like:

```
   [
        {
            attr1: value,
            attr2: value,
               ...
          },
        {
            attr1: value,
            attr2: value,
               ...
          }
    ]
```
Or, if there's no result, then `[]`

So, in both cases, the `result` type `[]map[string]interface{}` should work.

Am I missing something? If you have encountered a case where it doesn't work, could you please show the response data - so we can discuss how best to address it?

object_manager.go
saiprasannasastry: actually it is not necessary even when ipaddress is passed empty

connector.go
saiprasannasastry: @yuewko  shall i go ahead and resolve this comment atleast ? based on the discussion

yuewko: @saiprasannasastry , from our offline conversation, it sounds like that the reason the original code doesn't could be because: when we're doing a GET by reference, it returns a single map - and therefore []map[string]{} wouldn't work.

If that is the case, you might want to check for whether ref is used for query, if so, pass interface{}(or map[string]interface) to unmarshall, otherwise pass []map[string]interface{} like the original code.

Otherwise, I'm OK as implemented.

Thanks!

object_manager.go
```diff
@@ -336,33 +338,50 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
 }
 
 func (objMgr *ObjectManager) DeleteFixedAddress(ref string) (string, error) {
-        return objMgr.connector.DeleteObject(ref)
+	return objMgr.connector.DeleteObject(ref)
 }
 
-func (objMgr *ObjectManager) UpdateFixedAddress(fixedAddrRef string, matchclient string, macAddress string, vmID string, vmName string) (*FixedAddress, error) {
+// validate string in slice for match_client
+func validateStringinSlice(value string) bool {
```
chinmayb: s/validateStringinSlice/validateMatchClient

object_manager.go
```diff
@@ -336,33 +338,50 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
+	match_client := [5]string{"MAC_ADDRESS", "CLIENT_ID", "RESERVED", "CIRCUIT_ID", "REMOTE_ID"}
 
+	for _, a := range match_client {
```
chinmayb: s/a/val

object_manager.go
```diff
@@ -336,33 +338,50 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
 		updateFixedAddr.Ea = ea
 	}
-
-	if matchclient != "" {
-	updateFixedAddr.MatchClient = matchclient
+	if matchClient != "" {
+		if validateStringinSlice(matchClient) {
+			updateFixedAddr.MatchClient = matchClient
+		} else {
+			fmt.Printf("wrong value for match_client passed %s \n ", matchClient)
```
chinmayb: why not return an error here ?

object_manager.go
```diff
@@ -415,11 +434,13 @@ func (objMgr *ObjectManager) CreateHostRecord(enabledns bool, recordName string,
...
 	ea["VM ID"] = "N/A"
 	ea["VM Name"] = "N/A"
-	if vmID != "" &&vmName != "" {
+	if vmID != "" {
 		ea["VM ID"] = vmID
-		ea["VM Name"] = vmName
 	}
 
+	if vmName != "" {
+		ea["VM Name"] = vmName
```
chinmayb: assign a constant for `VM Name` and `VM ID`, and use it everywhere

objects_test.go
```diff
@@ -228,17 +228,20 @@ var _ = Describe("Objects", func() {
...
 			ipAddress := "25.0.7.59/24"
 			mac := "11:22:33:44:55:66"
+			matchClient := "MAC_ADDRESS"
 			fixedAddr := NewFixedAddress(FixedAddress{
 				NetviewName: netviewName,
 				Cidr:        cidr,
 				IPAddress:   ipAddress,
-				Mac:         mac})
+				Mac:         mac,
+				MatchClient: matchClient})
```
chinmayb: add a case for incorrect matchClient

saiprasannasastry: actually this in in objects_test.go , here we are checking if the objects are done or not.
And we have not written any test cases for Update calls in object_manager_test.go

object_manager.go
saiprasannasastry: ok

object_manager.go
saiprasannasastry: ok

object_manager.go
```diff
@@ -317,19 +336,51 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
+	return fixedAddr, err
+}
 
+func (objMgr *ObjectManager) DeleteFixedAddress(ref string) (string, error) {
+	return objMgr.connector.DeleteObject(ref)
+}
+
+// validation  for match_client
+func validateMatchClient(value string) bool {
+	match_client := [5]string{"MAC_ADDRESS", "CLIENT_ID", "RESERVED", "CIRCUIT_ID", "REMOTE_ID"}
```
yuewko: You should use a map here.

object_manager.go
```diff
@@ -317,19 +336,51 @@ func (objMgr *ObjectManager) GetFixedAddress(netview string, cidr string, ipAddr
...
 
+	ea := objMgr.getBasicEA(true)
 	if vmID != "" {
-		ea := objMgr.getBasicEA(true)
 		ea["VM ID"] = vmID
 		updateFixedAddr.Ea = ea
 	}
+	if vmName != "" {
+		ea["VM Name"] = vmName
+		updateFixedAddr.Ea = ea
```
yuewko: I am not sure about this. So, you are basically updating the Fix Address's EA with Basic EA (plus these updates to "VM ID" and "VM Name" EAs). So, what if the Fix Address already have EAs, would they get overwritten? Have you tested it?

object_manager.go
```diff
@@ -435,17 +489,22 @@ func (objMgr *ObjectManager) GetIpAddressFromHostRecord(host HostRecord) (string
...
+	ea := objMgr.getBasicEA(true)
 	if vmID != "" {
-		ea := objMgr.getBasicEA(true)
 		ea["VM ID"] = vmID
 		updateHostRecord.Ea = ea
 	}
 
+	if vmName != "" {
+		ea["VM Name"] = vmName
+		updateHostRecord.Ea = ea
```
yuewko: Same comment as I've made for Fixed Address.

object_manager.go
saiprasannasastry: yes it would get overwritten. vmId was already present , i did add vmName just to make sure attributes of vm would get covered

object_manager.go
saiprasannasastry: same comment,  vmId was already present , i did add vmName just to make sure attributes of vm would get covered

object_manager.go
yuewko: What I mean is,  here `ea` only contain EAs returned by `getBasicEA()`, plus the two you set here. What if the Fixed Address you are updating has ***other*** EAs that are set by the user, for example, manually in NIOS. Would those EAs get erased?

saiprasannasastry: Nope ,they won't get erased.and that has been tested

yuewko: OK. Sounds good! Thanks!

object_manager.go
yuewko: Since this is just a small array, let's just leave it the way it is then.

yuewko: Since it's just a small array, let's just leave it the way it is then.

PR #89, opened 2020-03-05
CRUD management for TXT record

ytsarev:


ytsarev:
@chinmayb @saiprasannasastry guys, any feedback ?

object_manager.go
```diff
@@ -562,6 +562,66 @@ func (objMgr *ObjectManager) DeleteCNAMERecord(ref string) (string, error) {
...
+}
+
+func (objMgr *ObjectManager) GetTXTRecordByRef(ref string) (*RecordTXT, error) {
+	recordTXT := NewRecordTXT(RecordTXT{})
+	err := objMgr.connector.GetObject(recordTXT, ref, &recordTXT)
+	return recordTXT, err
+}
+
+func (objMgr *ObjectManager) GetTXTRecord(name string) (*RecordTXT, error) {
+	if len(name) == 0 {
```
chinmayb:
```suggestion
	if name == "" {
	   return nil, fmt.Errorf("name can not be empty")
```

object_manager.go
```diff
@@ -562,6 +562,66 @@ func (objMgr *ObjectManager) DeleteCNAMERecord(ref string) (string, error) {
...
+
+	if err != nil || res == nil || len(res) == 0 {
+		return nil, err
+	}
+
+	return &res[0], nil
+}
+
+func (objMgr *ObjectManager) UpdateTXTRecord(recordname string, text string) (*RecordTXT, error) {
+	var res []RecordTXT
```
chinmayb:
```suggestion
	var res *RecordTXT
```
is it necessary for it be a slice ?

object_manager.go
```diff
@@ -562,6 +562,66 @@ func (objMgr *ObjectManager) DeleteCNAMERecord(ref string) (string, error) {
...
+
+	recordTXT := NewRecordTXT(RecordTXT{Name: recordname})
+
+	err := objMgr.connector.GetObject(recordTXT, "", &res)
+
+	res[0].Text = text
+
+	res[0].Zone = "" // otherwize infoblox API freaks out with ""text": "Field is not writable: zone""
```
chinmayb: by default it must be "" if you dont mention it, wondering why it needs to be explicitly mentioned ?

ytsarev: @chinmayb exactly,  if Zone is set or inherited from retrieved object then update will fail with the message from the comment

object_manager.go
ytsarev: @chinmayb thanks a lot, actually no need to. Updated.

object_manager.go
ytsarev: @chinmayb thanks a lot , updated

object_manager.go
chinmayb: got it, please update the description 
```suggestion
	res[0].Zone = "" //  set the Zone value to "" as its a non writable field
```

ytsarev: @chinmayb totally, thanks, done

ytsarev:
@chinmayb i've squashed the commits so ready for merge if you didn't spot anything else. Thanks a ton for in-depth review.

ytsarev:
@chinmayb I've added TTL specification support during record creation https://github.com/infobloxopen/infoblox-go-client/pull/89/commits/0dd42c77df20c2c8387c84968f75ed136235d239 , please check it out

ytsarev:
@chinmayb spotted the issue with GetTXTRecord during e2e testing, https://github.com/infobloxopen/infoblox-go-client/pull/89/commits/508141b434e53a619cc85c0a94045eb2ba0fb888 - here is the fix

object_manager.go
```diff
@@ -562,6 +562,68 @@ func (objMgr *ObjectManager) DeleteCNAMERecord(ref string) (string, error) {
...
+}
+
+func (objMgr *ObjectManager) UpdateTXTRecord(recordname string, text string) (*RecordTXT, error) {
+	var res []RecordTXT
+
+	recordTXT := NewRecordTXT(RecordTXT{Name: recordname})
+
+	err := objMgr.connector.GetObject(recordTXT, "", &res)
+
+	res[0].Text = text
```
chinmayb: its better to have len(res) == 0 check here before doing `res[0].Text`

object_manager.go
ytsarev: @chinmayb actually it is really required, see https://github.com/infobloxopen/infoblox-go-client/pull/89/commits/ba9fe18f76623d2336dedb9361e5c3f7bf7a5c8a .

object_manager.go
ytsarev: @chinmayb great point, please check https://github.com/infobloxopen/infoblox-go-client/pull/89/commits/1a033234161e60a3df814ffe1ed714154289d451

ytsarev:
@chinmayb merge ? :)

PR #96, opened 2020-05-19
New release

Anaghakh:
As a part of my internship project, I am working on enhancing the client.
We've taken Docker and Kubernetes Go-client libraries as our references.
As an initial commit, this would be specifically enhancing A Record subsequently we would be working on other records as well. 
There are a few changes committed in objects and object manager to align with our architecture.
With this new commit, we may have to think of releasing a newer version of Go-Client.
We are open to your suggestions.


saiprasannasastry:
@Anaghakh dont keep lines commented. instead delete all the comments.

PR #97, opened 2020-05-19
Enhancing the client

Anaghakh:
As a part of my internship project, I am working on enhancing the client.
We've taken Docker and Kubernetes Go-client libraries as our references.
As an initial commit, this would be specifically enhancing A Record subsequently we would be working on other records as well. 
There are a few changes committed in objects and object manager to align with our architecture.
With this new commit, we may have to think of releasing a newer version of Go-Client.
We are open to your suggestions.


record_a.go
```diff
@@ -0,0 +1,150 @@
...
+	GetARecord(recA RecordA) (*[]RecordA, error)
+	DeleteARecord(recA RecordA) (string, error)
+	UpdateARecord(recA RecordA) (*RecordA, error)
+}
+type Dhcpmember struct {
+	Ipv4addr string    `json:"ipv4addr,omitempty"`
+	Ipv6addr string    `json:"ipv6addr,omitempty"`
+	Name     string    `json:"name,omitempty"`
+}
+type Cloud_info struct {
```
chinmayb, line 19:
```suggestion
type CloudInfo struct {
```

record_a.go
```diff
@@ -0,0 +1,150 @@
...
+	_ "time"
+)
+
+type RecordAOperations interface {
+	CreateARecord(recA RecordA) (*RecordA, error)
+	GetARecord(recA RecordA) (*[]RecordA, error)
+	DeleteARecord(recA RecordA) (string, error)
+	UpdateARecord(recA RecordA) (*RecordA, error)
+}
+type Dhcpmember struct {
```
chinmayb, line 14:
```suggestion
type DHCPMember struct {
```

record_a.go
```diff
@@ -0,0 +1,150 @@
...
+	Creation_time    int    `json:"creation_time,omitempty"`
+	Comment          string `json:"comment,omitempty"`
+	Creator          string  `json:"creator,omitempty"`
+	Ddns_protected   bool    `json:"ddns_protected,omitempty"`
+	Dns_name         string  `json:"dns_name,omitempty"`
+	Forbid_reclamation bool  `json:"forbid_reclamation,omitempty"`
+	Reclaimable      bool     `json:"reclaimable,omitempty"`
+ 	Ttl              uint     `json:"ttl,omitempty"`
+	Use_ttl          bool     `json:"use_ttl,omitempty"`
+	//CloudInfo       Cloud_info `json:"cloud_info,omitempty"`
```
chinmayb, line 52: please review the commented if not needed

record_a.go
```diff
@@ -0,0 +1,150 @@
...
+	DeleteARecord(recA RecordA) (string, error)
+	UpdateARecord(recA RecordA) (*RecordA, error)
+}
+type Dhcpmember struct {
+	Ipv4addr string    `json:"ipv4addr,omitempty"`
+	Ipv6addr string    `json:"ipv6addr,omitempty"`
+	Name     string    `json:"name,omitempty"`
+}
+type Cloud_info struct {
+	Authority_type      string       `json:"authority_type,omitempty"`
```
chinmayb, line 20: please use camelcase instead of snake case

Anaghakh:
> Thanks for your contribution! Can we update the new README.MD with the new client ideas or how clients will use the v2 version of this client library and what are the problems we're solving as part of v2?

Yes, we have started working on it. We'll try to update it shortly.

record_a.go
Anaghakh: > please use camelcase instead of snake case

Okay, I will change that.
Thank you for the feedback!

record_a.go
```diff
@@ -0,0 +1,134 @@
...
+
+	var res []RecordA
+	recordA := NewRecordA(recA)
+	var err error
+	if len(recA.Ref)>0 {
+		err = objMgr.connector.GetObject(recordA, recA.Ref, &recordA)
+		res = append(res,*recordA)
+
+	} else {
+		recordA = NewRecordA(recA)
```
somashekhar, line 78: I think this is a redundant line of code @Anaghakh . Can you once verify if so.

Anaghakh: Yes, it was a redundant line. I've removed the line of code in both GET and DELETE. Thank you.

justinrush:
Looks like this PR solves what  I'm looking for in this issue:
https://github.com/infobloxopen/infoblox-go-client/issues/98

When will it be merged?

record_a.go
```diff
@@ -0,0 +1,132 @@
...
+func (objMgr *ObjectManager) DeleteARecord(recA RecordA) (string, error) {
+	var res []RecordA
+	recordName := NewRecordA(recA)
+	if len(recA.Ref) > 0 {
+		return  objMgr.connector.DeleteObject(recA.Ref)
+
+	} else {
+		err := objMgr.connector.GetObject(recordName, "", &res)
+		if err != nil || res == nil || len(res) == 0 {
+			return "Record doesn't exist", err
```
chinmayb, line 99:
```suggestion
			return "", err
```
This piece of information can be returned as part of `err`. 
`Record doesn't exist`

Anaghakh: Okay

PR #114, opened 2021-02-10
Change return value and type for UpdateTXTRecord. Fixes #113

### LINKED ISSUE #113: Update TXT Record - New object ref is not returned
https://github.com/infobloxopen/infoblox-go-client/blob/f5f00847ae7d426c9ddb42473ec161845edcd781/object_manager.go#L633-L639

When updating an Infoblox TXT record, the new ref is not returned, as the value is discarded when calling `connector.UpdateObject` (see above), and the old value (`&res[0]`) is the one returned. This causes a 404 at the next time the TXT record needs a getbyid run on it, as the actual ref/ID is now different.

See the below for an example where the update works (in this case, updating the host record), and returns the ref of the new object:

https://github.com/infobloxopen/infoblox-go-client/blob/f5f00847ae7d426c9ddb42473ec161845edcd781/object_manager.go#L519-L520

GNewbury1:
Fixes #113 by returning the reference to the new object when calling UpdateTXTRecord

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -21,7 +21,7 @@ type IBObjectManager interface {
 	CreateNetworkView(name string) (*NetworkView, error)
 	CreatePTRRecord(netview string, dnsview string, recordname string, cidr string, ipAddr string, ea EA) (*RecordPTR, error)
 	DeleteARecord(ref string) (string, error)
-	DeleteZoneAuth(ref string) (string, error) 
+	DeleteZoneAuth(ref string) (string, error)
 	DeleteCNAMERecord(ref string) (string, error)
 	DeleteFixedAddress(ref string) (string, error)
 	DeleteHostRecord(ref string) (string, error)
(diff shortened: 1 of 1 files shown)


object_manager.go
```diff
@@ -615,28 +615,28 @@ func (objMgr *ObjectManager) GetTXTRecord(name string) (*RecordTXT, error) {
...
-	_, err = objMgr.connector.UpdateObject(&res[0], res[0].Ref)
+	ref, err := objMgr.connector.UpdateObject(&res[0], res[0].Ref)
 
 	if err != nil || res == nil || len(res) == 0 {
-		return nil, err
+		return "", err
 	}
 
-	return &res[0], nil
+	return ref, nil
```
anagha-infoblox, line 639: I would suggest returning of pointer to the object by updating the reference of the object beforehand.

object_manager.go
```diff
@@ -615,28 +615,28 @@ func (objMgr *ObjectManager) GetTXTRecord(name string) (*RecordTXT, error) {
...
-		return nil, nil
+		return "", nil
 	}
 
 	res[0].Text = text
 
 	res[0].Zone = "" //  set the Zone value to "" as its a non writable field
 
-	_, err = objMgr.connector.UpdateObject(&res[0], res[0].Ref)
+	ref, err := objMgr.connector.UpdateObject(&res[0], res[0].Ref)
```
anagha-infoblox, line 633: The reference of the object can be updated after this line.
res[0].Ref=ref

somashekhar:
As a best practice I also support returning of an object rather a reference of it.

somashekhar:
Hi @GNewbury1, do you want to roll back this PR ?

somashekhar:
Closing this PR as of now.

PR #115, opened 2021-02-11
IPv6 Network

anagha-infoblox:
Extended support to perform CRUD operations on IPv6 Network.
Added a function to allocate an IPv6 network.


object_manager.go
```diff
@@ -441,6 +448,124 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
...
+	if err != nil {
+		return nil, err
+	}
+
+	for k, v := range updateEA {
+		res.Ea[k] = v
+	}
+
+	for k := range removeEA {
+		_, ok := res.Ea[k]
```
skudriavtsev, line 540: Minor improvement: 'found' instead of 'ok' seems to be more clear name.

object_manager.go
```diff
@@ -441,6 +448,124 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
...
+		if ok {
+			delete(res.Ea, k)
+		}
+	}
+
+	reference, err := objMgr.connector.UpdateObject(&res, ref)
+	res.Ref = reference
+	return &res, err
```
skudriavtsev, line 548: Canonical way is to check if err != nil and return object only in case if err == nil.

object_manager.go
```diff
@@ -441,6 +448,124 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
...
+	cidr = fmt.Sprintf("func:nextavailablenetwork:%s,%s,%d", cidr, netview, prefixLen)
+	ea := objMgr.getBasicEA(true)
+	networkReq := NewIPv6Network(netview, cidr, ea)
+	if name != "" {
+		networkReq.Ea["Network Name"] = name
+	}
+
+	ref, err := objMgr.connector.CreateObject(networkReq)
+	if err == nil && len(ref) > 0 {
+		network = BuildIPv6NetworkFromRef(ref)
```
skudriavtsev, line 563: Alternative way is to retrieve the object by its reference, for example using objMgr.connector.GetObject(). This will allow to avoid parsing object reference string for finding network view and CIDR by BuildIPv6NetworkFromRef() function.

objects.go
```diff
@@ -104,6 +104,25 @@ func NewNetwork(nw Network) *Network {
 	return &res
 }
 
+type IPv6Network struct {
```
skudriavtsev, line 107: Need to consider using existing object 'Network' for both v4 and v6. Most probably, this will require some discussion.

skudriavtsev (commented):
My comments seem to be minor at this point, but not addressing them in the nearest future may lead to some confusions and unclarities related to our API provided by this package. And even to some difficulties while using/extending the API in the future. Need to discuss all this.

object_manager_test.go
```diff
@@ -349,6 +351,151 @@ var _ = Describe("Object Manager", func() {
...
+		tenantID := "01234567890abcdef01234567890abcdef"
+		netviewName := "default_view"
+		cidr := "2001:db8:abcd:0012::0/64"
+		prefixLen := uint(65)
+		networkName := "private-net"
+		cidr1 := fmt.Sprintf("func:nextavailablenetwork:%s,%s,%d", cidr, netviewName, prefixLen)
+		fakeRefReturn := fmt.Sprintf("ipv6network/ZG5zLm5ldHdvcmskODkuMC4wLjAvMjQvMjU:%s/%s", cidr, netviewName)
+		anFakeConnector := &fakeConnector{
+			createObjectObj: NewIPv6Network(netviewName, cidr1, nil),
+			resultObject:    BuildIPv6NetworkFromRef(fakeRefReturn),
```
skudriavtsev, line 396: We are going to allocate network with prefix length 65 but here I see 'fakeRefReturn' which is based on 'cidr' with prefix length 64. Could you explain, why?

object_manager_test.go
```diff
@@ -349,6 +351,151 @@ var _ = Describe("Object Manager", func() {
...
+
+		objMgr := NewObjectManager(nwFakeConnector, cmpType, tenantID)
+
+		var actualRef *IPv6Network
+		var err error
+		It("should pass expected IPv6 Network Ref to getObject", func() {
+			actualRef, err = objMgr.GetIPv6NetworkWithRef(fakeRefReturn)
+		})
+		It("should return expected IPv6 Network record Ref", func() {
+			Expect(*actualRef).To(Equal(nwFakeConnector.resultObject.([]IPv6Network)[0]))
```
skudriavtsev, line 469: It also makes sense to check whether no more than one object is returned.

object_manager_test.go
```diff
@@ -349,6 +351,151 @@ var _ = Describe("Object Manager", func() {
...
+			Expect(err).To(BeNil())
+		})
+	})
+
+	Describe("Delete IPv6 Network", func() {
+		cmpType := "Docker"
+		tenantID := "01234567890abcdef01234567890abcdef"
+		netviewName := "default_view"
+		cidr := "2001%3Adb8%3Aabcd%3A0012%3A%3A0/64"
```
skudriavtsev, line 478: I think, this CIDR is not a valid one. Why '%3A' character sequence is used instead of ':'?

skudriavtsev (changes requested):
Test cases reviewed.

skudriavtsev:
One more thing. Please add negative test cases, only positive tests are not enough.

object_manager.go
```diff
@@ -19,14 +20,16 @@ type IBObjectManager interface {
 	CreateNetwork(netview string, cidr string, name string) (*Network, error)
 	CreateNetworkContainer(netview string, cidr string) (*NetworkContainer, error)
 	CreateNetworkView(name string) (*NetworkView, error)
+	CreateIPv6Network(netview string, cidr string, name string) (*Network, error)
```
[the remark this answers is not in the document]
skudriavtsev: It seems that you must return a reference to an object of IPv6Network type, not Network type which is actually IPv4.

object_manager.go
```diff
@@ -441,6 +448,127 @@ func (objMgr *ObjectManager) CreateEADefinition(eadef EADefinition) (*EADefiniti
...
+			ipv6Network.Cidr = cidr
+		}
+
+		if ea != nil && len(ea) > 0 {
+			ipv6Network.eaSearch = EASearch(ea)
+		}
+
+		err := objMgr.connector.GetObject(ipv6Network, "", &res)
+
+		if err != nil || res == nil || len(res) == 0 {
```
skudriavtsev, line 484: If err == nil but res == nil or len(res) == 0 as well, then the error returned will be nil also, and this is a problem. You should distinguish these cases, and in case when err == nil but we have an error by some other reason, you have to generate your own error message and return it.

skudriavtsev (changes requested):

skudriavtsev (approved):

somashekhar:
@anagha-infoblox  let us close this PR since an internal development is being done by you and will raise a new PR once that is tested.

PR #131, opened 2021-06-01
Use proxy env variables if explicit proxy not set

alexjfisher:
With this change, users can use the de facto standard HTTP(S)_PROXY environment variables instead of having to specify them manually. 
See https://golang.org/pkg/net/http/#ProxyFromEnvironment

I think using the environment variables is what you often get by [default](https://github.com/golang/go/blob/2e59cc5fb4e81fbd2ee9f662caa707c1138bf5ae/src/net/http/transport.go#L44)?? (sorry, I'm new to go)

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -133,6 +133,7 @@ func (whr *WapiHttpRequestor) Init(cfg TransportConfig) {
 			RootCAs: cfg.certPool,
 			Renegotiation:      tls.RenegotiateOnceAsClient},
 		MaxIdleConnsPerHost: cfg.HttpPoolConnections,
+		Proxy: http.ProxyFromEnvironment,
 	}
 
 	if cfg.ProxyUrl != nil {


somashekhar (commented):
Would recommend by default proxy should be taken from environment variable. Allow being overridden when explicitly given by an user through a client.
Also would suggest  `DefaultTransport` way of defining proxy for `ProxyFromEnvironment` [https://github.com/golang/go/blob/master/src/net/http/transport.go#L43](url)

alexjfisher:
@somashekhar Thanks for looking.

> Would recommend by default proxy should be taken from environment variable. Allow being overridden when explicitly given by an user through a client.

Unless I'm misunderstanding, I think that's exactly what this PR does. If `cfg.ProxyUrl` has been explicitly set, it'll be used, otherwise the environment variables will be used (if set obviously).

somashekhar:
> Unless I'm misunderstanding, I think that's exactly what this PR does. If `cfg.ProxyUrl` has been explicitly set, it'll be used, otherwise the environment variables will be used (if set obviously).

My suggestion was, let us initialize the `http.Transport`  with a default proxy from environment as `Proxy: ProxyFromEnvironment,` at [https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L131](url) rather than having it under an else loop. 

The way `DefaultTransport` [https://github.com/golang/go/blob/master/src/net/http/transport.go#L43](url) does it.

somashekhar:
> Unless I'm misunderstanding, I think that's exactly what this PR does. If `cfg.ProxyUrl` has been explicitly set, it'll be used, otherwise the environment variables will be used (if set obviously).

Suggesting to consider initializing  `http.Transport` with `Proxy: http.ProxyFromEnvironment` at https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L131 rather than having it under an else loop.

alexjfisher:
> > Unless I'm misunderstanding, I think that's exactly what this PR does. If `cfg.ProxyUrl` has been explicitly set, it'll be used, otherwise the environment variables will be used (if set obviously).
> 
> Suggesting to consider initializing `http.Transport` with `Proxy: http.ProxyFromEnvironment` at https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L131 rather than having it under an else loop.

Ah... understood.  I've updated the PR.  Hope that's what you meant! :)

NB. Untested. (I'll verify tomorrow and comment back.)

alexjfisher:
> NB. Untested. (I'll verify tomorrow and comment back.)

Code works fine like this too and is neater.  Thanks for the suggestion.

somashekhar (approved):
LGTM

skudriavtsev (approved):

alexjfisher:
@somashekhar Thanks for merging this!  No rush, but is there a timeframe for the next release that would include this, so I can create a PR on https://github.com/infobloxopen/terraform-provider-infoblox ?

Thanks.

somashekhar:
We are working on some of major features and would be including this particular go client changes along with them.
You can raise a PR if you are working on Terraform side code. If it is this particular changes only then not needed. We would make sure to rebase the code before release.

Issue #121, opened 2021-03-04
ObjectManager does not implement IBObjectManager


survivorbat:
Hey guys, I'm trying to create a REST API around InfoBlox and I'm attempting to use dependency injection using interfaces and structs so I can mock the IBObjectManager in my tests.

Now when I try to create an ObjectManager using:

```go
func GetInfoBloxObjectManager() *ibclient.IBObjectManager {
  // [...]
  return &ibclient.NewObjectManager(conn, "client", "")
}
```

My IDE shows me the error: 

`Type does not implement 'ibclient.IBObjectManager' need method: GetCNAMERecordByRef(ref string) (*RecordA, error) have method: GetCNAMERecordByRef(ref string) (*RecordCNAME, error)`

Is IBObjectManager not supposed to be the interface of the ObjectManager? Or am I doing something wrong here?

marcowork:
I had the same error +1!

survivorbat:
Edit:

Found another discrepancy:
`need method: GetZoneAuthByRef(ref string) (*ZoneAuth, error) have method: GetZoneAuthByRef(ref string) (ZoneAuth, error)`

marcowork:
@survivorbat, what version of infoblox are you using?

survivorbat:
Created #132

survivorbat:
This issue is resolved in the develop branch, I'll await release

PR #132, opened 2021-06-18
Ensure IBObjectManager is compatible with ObjectManager (#121)

survivorbat:
Related to #122 

The given interface does not adhere to the implementation (typos) and is unusable for dependency injection. This PR adds a compile-time check and updates the interface to be compatible with the struct.

Considering that the interface is unusable right now, I don't think this will cause any issues with backwards compatibility.

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -605,13 +605,13 @@ func (objMgr *ObjectManager) DeleteCNAMERecord(ref string) (string, error) {
 }
 
 // Creates TXT Record. Use TTL of 0 to inherit TTL from the Zone
-func (objMgr *ObjectManager) CreateTXTRecord(recordname string, text string, ttl int, dnsview string) (*RecordTXT, error) {
+func (objMgr *ObjectManager) CreateTXTRecord(recordname string, text string, ttl uint, dnsview string) (*RecordTXT, error) {
 
 	recordTXT := NewRecordTXT(RecordTXT{
 		View: dnsview,
 		Name: recordname,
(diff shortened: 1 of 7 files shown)


marcowork:
+1

object_manager.go
```diff
@@ -40,7 +43,7 @@ type IBObjectManager interface {
 	GetNetworkContainer(netview string, cidr string) (*NetworkContainer, error)
 	GetNetworkView(name string) (*NetworkView, error)
 	GetPTRRecordByRef(ref string) (*RecordPTR, error)
-	GetZoneAuthByRef(ref string) (*ZoneAuth, error)
+	GetZoneAuthByRef(ref string) (ZoneAuth, error)
```
somashekhar, line 46: We recommend to pass a pointer to an object rather than the object. Is there any specific reason for this change ?

somashekhar (changes requested):
We would recommend this PR to be raised on `develop` branch than on `master`. 

Also introducing line 11 code would need some additional changes to be done at `master` branch. The present PR would break the functionalities of `master` branch.

object_manager.go
survivorbat: This is how the ObjectManager was coded, I'm only updating the interface to adhere to that. I agree that it should be a pointer.

survivorbat:
I'll update it to develop

survivorbat:
This issue is fixed in the develop branch, so will add a tiny piece of code to that branch,

Issue #116, opened 2021-02-15
Feature Request: Ability to add comments


bmayfi3ld:
As the title says, I would like to add a comment to these records I'm creating. This may not work well with some of the "New" functions since it would break people's code, but it looks like it could be added to all the objects with the appropriate json tags.

somashekhar:
Hi @bmayfi3ld the enhancement to support comments for multiple records has been done and we are working on supporting new records with comment field. The Changes are in develop branch [https://github.com/infobloxopen/infoblox-go-client/tree/develop](url) . Please do use the same and provide the feedback.

somashekhar:
Now the feature is available at `master` branch. Will be available as part of `v2.0.0`

bmayfi3ld:
@somashekhar Thanks for getting to it, apologies I was not able to test it.

somashekhar:
> @somashekhar Thanks for getting to it, apologies I was not able to test it.

No problem😊. Was able to test it and merge.

Issue #110, opened 2020-11-16
Possibility to UPDATE a CNAME entry


deromemont:
related to https://github.com/terraform-providers/terraform-provider-infoblox/issues/36

Possibility to update a CNAME in the GO Library

It seems possible with the API

`curl -k -u admin:infoblox -H 'content-type: application/json' -X PUT "https://gridmaster/wapi/v2.11/record:cname/ZG5zLmJpbmRfY25hbWUkLl9kZWZhdWx0LmNvbS5pbmZvLmRlbW8uY2
5hbWV0ZXN0:cnametest.demo.info.com/default?_return_fields%2B=name&_return_as_object=1" -d
'{"name":"cname.demo.info.com"}'`

mmckane:
I would like to see this implemented as well.

somashekhar:
An update functionality has been added to CNAME Record. Please do use the same and provide feedback. We would work on to support it under Terraform.

somashekhar:
Now the feature is available at `master` branch. Will be available as part of `v2.0.0`

Issue #91, opened 2020-03-14
Support TTL configuration


ytsarev:
infoblox-go-client does not support TTL part of infoblox api:
![image](https://user-images.githubusercontent.com/518532/76682075-23329100-65f9-11ea-8d99-4634529e0d27.png)

My main use case is to modify TTL for specific TXT record https://github.com/AbsaOSS/ohmyglb/issues/61

ErwanMAS:
Can we have ttl support for A or CNAME or AAAA records ? MX NS ?

schlitzered:
from looking at the code, it seems only TXT records have implemented TTL functionality

somashekhar:
Now the feature is available at `master` branch. Will be available as part of `v2.0.0`

Support for MX and NS would be added soon.

Issue #86, opened 2020-02-15
IPv6 Support


nahun:
Enhancement request to support IPv6 in all applicable functions and objects and add AAAA support.

somashekhar:
Hi @nahun,
IPv6 support for DHCP objects has been done and we are working on supporting DNS objects. 
The changes are in develop branch [https://github.com/infobloxopen/infoblox-go-client/tree/develop](url) . Please do use the same and provide the feedback.

somashekhar:
We have added support for DNS objects as well.

Now the feature is available at `master` branch. Will be available as part of `v2.0.0`

PR #133, opened 2021-06-23
Deliverables-3

anagha-infoblox:
As part of phase 3 deliverables below are the features added:

- Support for AAAA record(CRUD operations)
- Update operation for A, CNAME and PTR records
- Get operation for CNAME and PTR records(both by reference and attributes)
- Added IPv6 support for PTR record
- EA and TTL support for A, AAAA, CNAME and PTR records

Solves #116 #110 #91 #86 #82

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -7,24 +7,27 @@ import (
 	"net"
 	"net/url"
 	"regexp"
+	"strings"
 )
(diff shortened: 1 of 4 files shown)


somashekhar (approved):
LGTM

PR #135, opened 2021-06-29
Add compile-time interface check

survivorbat:
Adds a compile-time check to the interface to make sure it stays in sync with the actual implementation.


somashekhar (approved):
LGTM.
@anagha-infoblox @skudriavtsev can you have a look.

PR #166, opened 2022-04-07
Create auto-generated objects.go file

achernevskii:
* ObjectType field of the IBObject is replaced with the ObjectType
  method for each child object.

* A lot of simple constructors in the objects.go file are removed. To
  create a struct instance one can use a struct literal.

* Removed Ipv4Addr/Ipv6Addr fields from the HostRecord object.

* Removed one unit-test for the HostRecord object.

* All other unit tests, and ObjectManager methods are changed in
  accordance to updated in the objects.go file.

* The IBObjectManager interface is not changed.


achernevskii:
Created a different PR for this change: #171

PR #171, opened 2022-04-20
Create auto-generated objects.go file

achernevskii:
# Issue/Feature description

* Structs for WAPI objects should be auto-generated to simplify the go-client usage and development.

# How it was fixed/implemented

* ObjectType field of the IBObject is replaced with the ObjectType
  method for each child object.

* Override ReturnFields method for each struct. By default, it will return
  a set of standard fields for each respective struct.

* Removed Ipv4Addr/Ipv6Addr fields from the HostRecord object.

* Updated existing object manager unit tests.

* The IBObjectManager interface is not changed.

# Tests

* Created a E2E connector facade, that will remove all objects created by e2e test after the test is done.

* Created a set of E2E tests to check crud operations on the generated object structs.

# Reviewers

@skudriavtsev


achernevskii:
This one is closed, because new PR #178  with additional e2e tests and bugfixes is open.

PR #180, opened 2022-08-05
Add paging get object method

skalpin:
In my other implementation I am using generics, however I wasn't prepared to upgrade this project to go 1.18 so I went with interface{} instead.

I don't think the json.Unmarshall call will create the correct type, but I wanted to offer this up.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -389,6 +389,58 @@ func (c *Connector) GetObject(
 	return
 }
 
+func (c *Connector) PagingGetObject(
+	obj IBObject, ref string,
+	queryParams map[string]string, res *[]interface{}) (err error) {
+
+	pagingResponse := pagingResponseStruct{
+		NextPageId: "",
+		Result:     *res,
+	}
+
+	//copy query params and update them
+	queryParamsCopy := map[string]string{}
+	for k, v := range queryParams {
+		queryParamsCopy[k] = v
+	}
+
+	queryParamsCopy["_return_as_object"] = "1"
+	queryParamsCopy["_paging"] = "1"
+
+	var pagingResult []interface{}
+
+	err = c.GetObject(obj, "", NewQueryParams(false, queryParamsCopy), &pagingResponse)
+	if err != nil {
+		return fmt.Errorf("could not fetch object: %s", err)
+	} else {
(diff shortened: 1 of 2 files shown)


hemanthKa677 (commented):
Hi @skalpin , Can you please provide us example testing this? Because usual get object takes parameter with res interface{}, which all operations do work with that. And please comment us how PagingGetObject consider the parameters.

hemanthKa677 (commented):
Hi @skalpin , Can you please respond, there are some changes required here?

1. Please declare PagingGetObject method in IBConnector interface{}
2. Please share us example argument res with type *[]interface

skalpin:
@hemanthKa677, I am sorry I missed this notification. I'll take a look at it today.

skalpin:
@hemanthKa677 I want this to work exactly like the existing GetObject for consumers with Infoblox zones that have > 1000 records. The problem is, if you call GetObject and your result query has > 1000 records, the Infoblox api returns an error. PagingGetObject will loop through the results until it can return _all_ of the records.

I've been using this https://github.com/GeneralMills/external-dns/blob/master/provider/infoblox/paging_ibclient.go and raised a PR to the External-DNS project https://github.com/kubernetes-sigs/external-dns/pull/2996

I think it would be best to include the change in the go client, however I wrote this using generics which weren't introduced until golang 1.18. There is a separate PR https://github.com/infobloxopen/infoblox-go-client/pull/182

I don't think the implementation in this PR will work properly because I can't declare an array with the correct type to store the paging results in. (or if I can, I don't know how)

Do you think it's possible to upgrade this to use golang 1.18?

connector.go
```diff
@@ -389,6 +389,58 @@ func (c *Connector) GetObject(
 	return
 }
 
+func (c *Connector) PagingGetObject(
+	obj IBObject, ref string,
+	queryParams map[string]string, res *[]interface{}) (err error) {
+
```
hemanthKa677, line 395: Hi @skalpin , Can you please say how you tested this code with these attributes? Was the code working for you?

skalpin: @hemanthKa677 I haven't tested this code. The implementation I am using, and have tested is using golang 1.18.

PR #186, opened 2022-09-12
Add useful fields for Authoritative Zone resource

### LINKED ISSUE #185: Add useful fields for Authoritative Zone resource
### Context

When creating an Authoritative Zone in DNS using the current client, only the desired FQDN is accepted as an argument. This does not allow a user to set a Nameserver Group, or the TTLs for SOA records, for instance.

```go
CreateZoneAuth(fqdn string, ea EA) (*ZoneAuth, error)
```

### Proposed Changes

Add the following fields for Authoritative Zone resources. These are **already present in the Infoblox NIOS API**, they just aren't exposed here in the Go Client.

* `ns_group`
* `restart_if_needed`
* `comment`
* `soa_default_ttl`
* `soa_expire`
* `soa_negative_ttl`
* `soa_refresh`
* `soa_retry`

### Related Work

With these fields exposed in the Infoblox Go Client, we can then add an Authoritative Zone resource to the [Infoblox Terraform Provider](https://github.com/infobloxopen/terraform-provider-infoblox).

(Related Issue and PR coming soon)

bsmithtm:
Fixes #185.

This adds the following fields for the `ZoneAuth` resource, so users can set a Nameserver Group or SOA TTLs when creating a new zone.

* `ns_group`
* `restart_if_needed`
* `comment`
* `soa_default_ttl`
* `soa_expire`
* `soa_negative_ttl`
* `soa_refresh`
* `soa_retry`

These are already present in the Infoblox NIOS API; this change exposes them here in the Go Client.

### DIFF
--- a/object_manager.go
+++ b/object_manager.go
@@ -15,7 +15,7 @@ type IBObjectManager interface {
 	AllocateNetworkContainer(netview string, cidr string, isIPv6 bool, prefixLen uint, comment string, eas EA) (netContainer *NetworkContainer, err error)
 	CreateARecord(netView string, dnsView string, name string, cidr string, ipAddr string, ttl uint32, useTTL bool, comment string, ea EA) (*RecordA, error)
 	CreateAAAARecord(netView string, dnsView string, recordName string, cidr string, ipAddr string, useTtl bool, ttl uint32, comment string, eas EA) (*RecordAAAA, error)
-	CreateZoneAuth(fqdn string, nsGroup string, restartIfNeeded bool, comment string, SoaDefaultTtl int, SoaExpire int, SoaNegativeTtl int, SoaRefresh int, SoaRetry int, ea EA) (*ZoneAuth, error)
+	CreateZoneAuth(dnsview string, fqdn string, nsGroup string, restartIfNeeded bool, comment string, soaDefaultTtl int, soaExpire int, soaNegativeTtl int, soaRefresh int, soaRetry int, zoneFormat string, ea EA) (*ZoneAuth, error)
 	CreateCNAMERecord(dnsview string, canonical string, recordname string, useTtl bool, ttl uint32, comment string, eas EA) (*RecordCNAME, error)
 	CreateDefaultNetviews(globalNetview string, localNetview string) (globalNetviewRef string, localNetviewRef string, err error)
 	CreateEADefinition(eadef EADefinition) (*EADefinition, error)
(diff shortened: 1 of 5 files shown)


hemanthKa677 (commented):
Looks good to me

object_manager.go
```diff
@@ -15,7 +15,7 @@ type IBObjectManager interface {
 	AllocateNetworkContainer(netview string, cidr string, isIPv6 bool, prefixLen uint, comment string, eas EA) (netContainer *NetworkContainer, err error)
 	CreateARecord(netView string, dnsView string, name string, cidr string, ipAddr string, ttl uint32, useTTL bool, comment string, ea EA) (*RecordA, error)
 	CreateAAAARecord(netView string, dnsView string, recordName string, cidr string, ipAddr string, useTtl bool, ttl uint32, comment string, eas EA) (*RecordAAAA, error)
-	CreateZoneAuth(fqdn string, ea EA) (*ZoneAuth, error)
+	CreateZoneAuth(fqdn string, nsGroup string, restartIfNeeded bool, comment string, SoaDefaultTtl int, SoaExpire int, SoaNegativeTtl int, SoaRefresh int, SoaRetry int, ea EA) (*ZoneAuth, error)
```
skudriavtsev, line 18: Please, unify the format of the names. Some of them start from capital letter, some are not. I think, all of them should start from a small letter. Just for consistency.

object_manager_zone_auth.go
```diff
@@ -0,0 +1,50 @@
+package ibclient
+
+import "fmt"
+
+func (objMgr *ObjectManager) CreateZoneAuth(
+	fqdn string,
```
skudriavtsev, line 6: The list of fields must be extended at least by 'view' (DNS view), and preferably by 'zone_format' (possible values in WAPI: FORWARD, IPV4, IPV6), with appropriate busyness-logic changes.

objects.go
```diff
@@ -824,17 +824,29 @@ func NewRecordTXT(
...
-	IBBase `json:"-"`
-	Ref    string `json:"_ref,omitempty"`
-	Fqdn   string `json:"fqdn,omitempty"`
-	View   string `json:"view,omitempty"`
-	Ea     EA     `json:"extattrs"`
+	IBBase          `json:"-"`
+	Ref             string `json:"_ref,omitempty"`
+	Fqdn            string `json:"fqdn,omitempty"`
+	View            string `json:"view,omitempty"`
+	NsGroup         string `json:"ns_group,omitempty"`
```
skudriavtsev, line 831: What about setting no NS groups for this zone (in UI: 'None' for 'Name Servers')? 'omitempty' option can prevent this.

skudriavtsev (changes requested):

object_manager.go
bsmithtm: I've set all the field names to start with lower case

object_manager_zone_auth.go
bsmithtm: I've added both DNS view and zone_format to the object and the applicable functions

objects.go
bsmithtm: If I understand correctly, you mean a case where the end user would like to send empty string "" as the `ns_group` to explicitly set this field to nothing; if we omit this from marshaling when it is empty, on the Infoblox side it will attempt to set the default NS Group instead. Is that correct?

bsmithtm:
Hi @skudriavtsev thanks for the review and feedback! I've made most of the requested changes:

* Make the case of the parameter names in object_manager consistent (all start lower case now)
* Add `view` and `zone_format` to the ZoneAuth object and applicable functions
* Make an `UpdateZoneAuth()` function to allow updating fields that are allowed in the WAPI (this is a request from the PR for the Terraform Provider)

If we can resolve the last question about `omitempty` on `ns_group`, I think all requests will be handled. Thanks!

PR #191, opened 2022-10-18
Added latest version fetching functionality

achernevskii:
* If no version is provided for the connector, NewConnector init function will fetch WAPI schema, with all the supported versions, and use the latest version for connector's WAPI calls.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -100,68 +100,58 @@ type IBConnector interface {
 }
 
 type Connector struct {
-	hostCfg        HostConfig
+	hostCfg        *HostConfig
 	authCfg        AuthConfig
 	transportCfg   TransportConfig
 	requestBuilder HttpRequestBuilder
 	requestor      HttpRequestor
 }
 
-func getLatestVersion(hostCfg HostConfig, authCfg AuthConfig, transportCfg TransportConfig) (string, error) {
-	schemaUrl := fmt.Sprintf("https://%s/wapi/v1.0/?_schema", hostCfg.Host)
-	req, err := http.NewRequest(http.MethodGet, schemaUrl, nil)
+// setLatestVersion will fetch latest WAPI version from provided instance, and set this version as default for
+// connector WAPI calls. Fields hostCfg, authCfg, requestor should be configured before calling this method.
+func (c *Connector) setLatestVersion() error {
+	schemaUrl := url.URL{
+		Scheme:   "https",
+		Host:     c.hostCfg.Host + ":" + c.hostCfg.Port,
+		Path:     fmt.Sprintf("wapi/v%s/", c.hostCfg.Version),
+		RawQuery: "_schema",
+	}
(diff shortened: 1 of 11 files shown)


connector.go
```diff
@@ -104,6 +107,63 @@ type Connector struct {
 	requestor      HttpRequestor
 }
 
+func getLatestVersion(hostCfg HostConfig, authCfg AuthConfig, transportCfg TransportConfig) (string, error) {
+	schemaUrl := fmt.Sprintf("https://%s/wapi/v1.0/?_schema", hostCfg.Host)
```
skudriavtsev, line 111: We must take into account hostCfg.Port as well.

connector.go
```diff
@@ -104,6 +107,63 @@ type Connector struct {
...
+	req, err := http.NewRequest(http.MethodGet, schemaUrl, nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.SetBasicAuth(authCfg.Username, authCfg.Password)
```
skudriavtsev, line 117: What will happen in the case when we do not use user's name and password at all? (certificate-based client's authentication)

connector.go
```diff
@@ -104,6 +107,63 @@ type Connector struct {
...
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: !transportCfg.SslVerify,
+		},
+	}
+
+	client := &http.Client{
+		Timeout:   time.Second * 10,
```
skudriavtsev, line 126: We should not use hard-coded timeout values whenever possible. Please use some user-specified option (CLI option, environment variable, etc), either already existing or introduced in this PR.

connector.go
```diff
@@ -104,6 +107,63 @@ type Connector struct {
...
+		Transport: tr,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != 200 {
```
skudriavtsev, line 134: please use http.StatusOk instead

connector.go
```diff
@@ -104,6 +107,63 @@ type Connector struct {
...
+		log.Printf("Non OK Response: %v", resp)
```
skudriavtsev, line 135: This form of messages is appropriate for 'debug' level, but for other cases there should be more formal and more detailed message (ex. including the text from resp.Status, maybe  along with response's body content)

skudriavtsev (changes requested):

connector.go
achernevskii: Updated the logic. Now client will use the requestor object, which uses client certs, and if username/password are configured, I will add them to the request.

connector.go
achernevskii: Updated the logic. Now I use the requestor object, which handles incorrect HTTP status codes.

connector.go
achernevskii: Updated the logic. Now I use the requestor object, which handles incorrect HTTP status codes.

PR #195, opened 2022-11-22
Make certPool public field of TransportConfig

MartinWeindel:
All fields of the `TransportConfig` are public, but the certPool field is not.
It should not be required to provide a cert pool from a file, but just by providing an existing CertPool object.


hemanthKa677 (commented):
Looks good to me

PR #197, opened 2023-01-13
Updates README file in configuration

hemanthKa677:
This PR updates the README file with seperate config's for hostConfig and authConfig(either with username and password or certificates)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,10 +45,10 @@ This library is compatible with Go 1.2+
    		    Port:     "PORT",
    	    }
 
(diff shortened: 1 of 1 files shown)


skudriavtsev:
The goal of this PR is to renew the example in the README.md document. But this PR does not do this: the example is not even compilable. Please fix this and make sure that it works fine (when proper server parameters are used), and do 'go fmt' against this example to fix the formatting before putting it into README.md document.

skudriavtsev (approved):

PR #198, opened 2023-01-19
Add RecordMX support in go-client

hemanthKa677:
This will add the CreateMXRecord , GetMXRecord, GetMXRecordByRef and DeleteMXRecord operations to the infoblox-go-client, so that terraform can also support for the MXRecord.


object_manager_mx_record.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, NewNotFoundError(
+			fmt.Sprintf(
+				"MX record with name '%s' in DNS view '%s' is not found",
+				fqdn, dnsview))
+	}
+
+	return &res[0], err
+}
+
+func (objMgr *ObjectManager) UpdateMXRecord(
```
skudriavtsev, line 71: This method is not a part of IBObjectManager interface's signature. Please fix this.

skudriavtsev: And again, do not forget about 'ttl' field.

object_manager_mx_record.go
```diff
@@ -0,0 +1,104 @@
...
+		dnsView:  dnsview,
+		Fqdn:     fqdn,
+		MX:       mx,
+		Priority: priority,
+		Comment:  comment,
+		Ea:       eas,
+	})
+
+	recordMx.Ref = ref
+
```
skudriavtsev, line 90: You should add some validation here. In particular, for 'priority' (it must be >0), 'ttl' (the same), 'dnsView' (this is not allowed to be updated), "mx" (validate as a domain name, up to you).

skudriavtsev (changes requested):

object_manager_mx_record.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
+	recordMX := NewRecordMX(RecordMX{})
+	err := objMgr.connector.GetObject(recordMX, ref, NewQueryParams(false, nil), &recordMX)
+
+	return recordMX, err
+}
+
+func (objMgr *ObjectManager) GetMXRecord(dnsview string, fqdn string) (*RecordMX, error) {
```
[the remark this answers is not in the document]
hemanthKa677: Sergey, Do you mean, that dnsview to be written here as dnsView?

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
+	dnsView string,
 	fqdn string,
 	mx string,
+	ttl uint32,
+	useTtl bool,
 	comment string,
 	priority int,
 	eas EA) (*RecordMX, error) {
 
+	res, _ := objMgr.GetMXRecordByRef(ref)
```
skudriavtsev, line 81: Why the error is ignored here? This is quite an important place to stop if something bad happens.

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
 	if err != nil {
 		return nil, err
-	} else if res == nil || len(res) == 0 {
-		return nil, NewNotFoundError(
-			fmt.Sprintf(
-				"MX record with name '%s' in DNS view '%s' is not found",
-				fqdn, dnsview))
 	}
 
 	return &res[0], err
```
skudriavtsev, line 67: In case of an error, this place cannot be reached. Thus, either the code above (which was removed) should be returned back, or the line should become 'return &res[0], nil'.

Removing the code above makes it harder for a developer to figure out (not obvious) how to distinguish between 'not found' error and other error types in the code, which uses GetMXRecord() method.

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
+	if dnsView != res.View {
+		return nil, fmt.Errorf("changing dns_view after resource creation is not allowed")
+	}
+
+	if priority < 0 {
+		return nil, fmt.Errorf("priority must be greater than zero")
+	}
+
+	if mx == "" {
+		return nil, fmt.Errorf("mx must not be empty")
```
skudriavtsev, line 92: 'mail_exchanger' field must not be empty

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
 	eas EA) (*RecordMX, error) {
 
+	res, _ := objMgr.GetMXRecordByRef(ref)
+
+	if dnsView != res.View {
+		return nil, fmt.Errorf("changing dns_view after resource creation is not allowed")
+	}
+
+	if priority < 0 {
+		return nil, fmt.Errorf("priority must be greater than zero")
```
skudriavtsev, line 88: "... must not be a negative number"

In the sense that '0' (zero) is also a valid value. By the way, this field is called not 'priority' in WAPI, but 'preference'.

objects.go
```diff
@@ -799,18 +799,20 @@ func NewHostRecord(
 type RecordMX struct {
 	IBBase   `json:"-"`
 	Ref      string `json:"_ref,omitempty"`
-	dnsView  string `json:"view,omitempty"`
+	View     string `json:"view,omitempty"`
 	Fqdn     string `json:"name,omitempty"`
 	MX       string `json:"mail_exchanger,omitempty"`
-	Priority int    `json:"preference,omitempty"`
+	Priority int
```
skudriavtsev, line 805: There must be a mapping to a field's name from WAPI. 'preference', I guess.

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
-		"view": dnsview,
+		"view": dnsView,
 		"name": fqdn,
 	}
 	queryParams := NewQueryParams(false, sf)
 	err := objMgr.connector.GetObject(recordMX, "", queryParams, &res)
 
 	if err != nil {
 		return nil, err
-	} else if res == nil || len(res) == 0 {
```
skudriavtsev: There was a misunderstanding between us. I did mean that 'else' keyword is redundant here, but the 'if' clause is still necessary. Please, return this back, but without 'else' keyword.

object_manager_mx_record.go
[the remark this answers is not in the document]
skudriavtsev: Yes, indeed. It is not necessary but is a good thing to do. Thanks for doing this.

object_manager_mx_record.go
```diff
@@ -41,47 +45,59 @@ func (objMgr *ObjectManager) GetMXRecordByRef(ref string) (*RecordMX, error) {
...
+	ttl uint32,
+	useTtl bool,
 	comment string,
 	priority int,
 	eas EA) (*RecordMX, error) {
 
+	res, _ := objMgr.GetMXRecordByRef(ref)
+
+	if dnsView != res.View {
+		return nil, fmt.Errorf("changing dns_view after resource creation is not allowed")
```
skudriavtsev, line 84: "... after creating an object...". 'resource' is from Terraform, WAPI deals with 'object'.

skudriavtsev (changes requested):

object_manager_mx_record.go
hemanthKa677: Sorry, Can you please elaborate this one thing , Sergey?

objects_test.go
```diff
@@ -603,6 +603,39 @@ var _ = Describe("Objects", func() {
...
+				Ttl:      ttl,
+				UseTtl:   useTtl,
+				Comment:  comment,
+				Ea:       eas,
+			})
+
+			It("should set fields correctly", func() {
+				Expect(rm.Fqdn).To(Equal(fqdn))
+				Expect(rm.MX).To(Equal(mx))
+				Expect(rm.View).To(Equal(dnsView))
```
skudriavtsev, line 630: What about other fields of 'rm' structure? I mean: Priority, Ttl, UseTtl, Comment, Ea.

object_manager_mx_record_test.go
```diff
@@ -0,0 +1,183 @@
+package ibclient
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Object Manager: MX-record", func() {
```
skudriavtsev, line 10: GetMXRecord() and GetMXRecordByRef() are not tested.

skudriavtsev (changes requested):

skudriavtsev:
replaced by https://github.com/infobloxopen/infoblox-go-client/pull/203

PR #200, opened 2023-02-01
Add RecordSRV support in go-client

hemanthKa677:
This PR contains commits to support all the Create, Read, Update and Delete Operations related to SRV record. These commits have methods like CreateSRVRecord, GetSRVRecord , UpdateSRVRecord, etc. This PR will also work for terraform plugin to support SRV Record


object_manager_srv-record.go
```diff
@@ -0,0 +1,151 @@
...
+
+	nw_ref, err := objMgr.connector.UpdateObject(recordSRV, ref)
+
+	if err != nil {
+		return nil, err
+	}
+
+	recordSRV.Ref = nw_ref
+
+	return recordSRV, err
```
skudriavtsev, line 146: Returning 'err' is pointless here, because if err != nil then we cannot reach this line.

skudriavtsev (changes requested):

skudriavtsev (changes requested):

skudriavtsev:
replaced by https://github.com/infobloxopen/infoblox-go-client/pull/203

PR #216, opened 2023-12-18
Update HostConfig with Scheme

smalisetti-infoblox:
Updates HostConfig with `Scheme` and use default "https" if Scheme is not "http". 
No breaking changes.

Updates example in Readme to demonstrate that Host is configurable with Scheme.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -28,6 +28,7 @@ type AuthConfig struct {
 }
 
 type HostConfig struct {
+	Scheme  string
 	Host    string
 	Version string
 	Port    string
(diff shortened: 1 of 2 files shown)


connector.go
```diff
@@ -250,8 +251,12 @@ func (wrb *WapiRequestBuilder) BuildUrl(t RequestType, objType string, ref strin
 		qry = vals.Encode()
 	}
 
+	scheme := "https"
```
pnagaraj80, line 254: Can we add validation to not support other schemes?

smalisetti-infoblox: making a change to just allow https or http

pnagaraj80 (approved):

hemanthKa677 (commented):
Changes looks good. 
But I found an error, while testing with my NIOS grid url.
If someone has `http` working url of NIOS grid, Could you please share?

pnagaraj80:
@hemanthKa677 http will be used only internally between 2 services in CSP backend. Basically Federated IPAM and RPC server in the below diagram. 

image


https://docs.google.com/document/d/1oN74COstNtehZjgUKxwvMvRznj2sJP_X2Pf-kbpov1A/edit#heading=h.kcrm8x1kk5la

PR #218, opened 2024-01-10
Introduce request builder with headers

smalisetti-infoblox:
the new type can be used like the following

``` go
header := make(http.Header)
header.Add("x", "1")
header.Add("y", "2")

wrb, _ := NewWapiRequestBuilder(hostCfg, authCfg)
wrbh, err := NewWapiRequestBuilderWithHeaders(wrb, header)

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -7,10 +7,29 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
+
+func TestHeaderRequestBuilder_BuildRequest(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: connector_test.go [hunk 1]
```diff
@@ -7,11 +7,30 @@ import (
...
 	"net/url"
 	"strings"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+func TestHeaderRequestBuilder_BuildRequest(t *testing.T) {
+	header := make(http.Header)
```
hemanthKa677, line 17: Could you please add as Describe function similar to other Describe functions?

hemanthKa677 (commented):
test case is working fine. Looks good for me.

PR #219, opened 2024-02-02
Move test files to a separate package

ylongca:
- This is to prevent the test dependencies being downloaded by downstream dependencies.


ibclient_test/connector_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
```
pnagaraj80, line 4: This file is still part of ibclient package. Is this expected?

ibclient_test/objects_test.go
```diff
@@ -3,27 +3,29 @@ package ibclient
 import (
```
pnagaraj80, line 3: This file is still part of ibclient package. Is this expected?

daniel-garcia:
how does this affect unit test coverage reports?

ylongca:
> how does this affect unit test coverage reports?

Unit test coverage cannot be reported because of this change.

hemanthKa677 (commented):
Hi @ylongca, Is it really necessary to move test files to seperate package? Could you please confirm on this?
I am not sure , because dependenices will be anyway downloaded from go.mod file like ginkgo dependency for testing.

daniel-garcia:
For now, lets create a branch that doesn't include the tests.

ylongca:
> For now, lets create a branch that doesn't include the tests.

done.
https://github.com/infobloxopen/infoblox-go-client/tree/remove-unit-tests

PR #221, opened 2024-02-23
Remove omitempty json tag from dns/dhcp struct

akshay8:
* Remove `omitempty` json tag from DNS and DHCP struct
* As in request build method, json marshal will remove zero values of the field. But EnableDns/EnableDhcp is required in the body.


hemanthKa677 (approved):
LGTM

PR #225, opened 2024-03-15
Added generic function for search by alt id.

sdas-infoblox:


object_manager.go
```diff
@@ -301,3 +556,93 @@ func (objMgr *ObjectManager) UpdateZoneDelegated(ref string, delegate_to []NameS
...
+
+// validateObjByInternalId validates the object by comparing the given internal with the object's internal id
+func validateObjByInternalId(res interface{}, internalId, eaNameForInternalId string) (bool, error) {
+	var success bool
+	if res == nil {
+		return success, nil
+	} else if strings.TrimSpace(internalId) == "" {
+		// return object if internal id is empty
+		success = true
+		return success, nil
```
hemanthKa677, line 617: why it is returned true here

hemanthKa677 (commented):
besides that comment, looks good to me

object_manager.go
sdas-infoblox: We should return the return response received when fetched by ref and the internal id is empty. I should have updated the comment to `return true if response is not nil and internal id is empty`.

hemanthKa677 (approved):
LGTM

PR #231, opened 2024-06-04
Feature/fwd zone

Aish-sp:
Adding support for forward zone.


JkhatriInfobox (approved):
Initial development for supporting FWD Zone

PR #233, opened 2024-06-18
[FIX] change GetForwardZone function parameter to queryParams

Aish-sp:


JkhatriInfobox (approved):
Changes w.r.t queryparam to support proxy looks good

Issue #147, opened 2021-08-18
Segfault due to missing error check with bad HostConfig settings


claytonpeters:
In [connector.go, line 264](https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L264), a call to `BuildRequest` is made, but `err` is never checked before `req` is passed to `SendRequest` on line 265. In the event `BuildRequest` fails, which is possible if the `Connector`'s `HostConfig` object has an invalid `Host` attribute (for example, if it's accidentally specified as a URL rather than an IP/hostname) then `req` is `nil`.

When this is passed to `SendRequest`, it is blindly passed to the `Do` method of `net/http.Client` which then segfaults.

There should probably be some error checking here to fail more gracefully and/or some validation of the contents of `HostConfig` in `Init` on [line 174](https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L174)

rhossili:
@claytonpeters 

IMO, there are two separate issues here.

1) Totally agree that there should be a check for error returned by `BuildRequest` before calling `SendRequest` [here](https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L265).

2) The above won't address the example case of an invalid `Host` attribute. As you suggested, this can be addressed by validating `HostConfig` in `Init`. In addition, I noticed that there's a call to `ValidateConnector` [here](https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L405) that is supposed to perform a validation by actually making a connection. However, the implementation has been commented out (see https://github.com/infobloxopen/infoblox-go-client/blob/master/connector.go#L378-L384). Perhaps we can resurrect that?

PR #241, opened 2024-08-20
connector: restore ValidateConnector

rascalking:
It looks like this has been silently disabled for three years.  Reenabled and updated to the current `GetObject` signature.

I didn't see any unit tests that exercise it, I can add some if needed.


rascalking:
It looks like this might help address #147.

PR #243, opened 2024-09-17
Fix for NPA-165

Chaithra001:
Fix for NPA-165 : - EAs are not retrieved by zone_delegated datasource


JkhatriInfobox (approved):
Changes looks good to me for Bug fix

PR #266, opened 2024-12-13
Added Get object by reference for DTC pools and servers, added GetAll…

Chaithra001:
Added  GetDtcPoolByRef  and GetDtcServerByRef  for DTC pools and servers respectively in IBObjectManager interface.
Added GetAllDtcServer and GetDtcServer functions with test cases.
Removed updateTopologyReference() function.
Replaced updateTopologyReference() with getTopology() due to similar functionality in create and update function for dtc pools


object_manager_dtc_server.go
Aish-sp, line 1: What happens if you pass empty string for name and host in GetDtcServer()?
Validation of parameters is required.

Aish-sp (changes requested):

object_manager_dtc_server.go
Chaithra001: Added error handling in the below commit

Aish-sp (approved):
LGTM

PR #269, opened 2025-01-07
Updating error message in updateServerReferences and getMonitorReference

Chaithra001:
Updating error message in updateServerReferences and getMonitorReference


Aish-sp (approved):
LGTM

PR #272, opened 2025-01-17
updating pool marshalJson function to support unsetting monitors, consolidated_monitors and servers fields

Aish-sp:


Chaithra001 (approved):
LGTM

PR #273, opened 2025-01-21
NPA-300, NPA-301 fix

Aish-sp:
1. updating error messages.
2. condition to unset consolidated_monitors field when consolidated_monitors is empty slice and auto_consolidated_monitors is set to false


Chaithra001 (approved):
LGTM

PR #288, opened 2025-03-28
network range implementation and test cases and

Chaithra001:
Network range implementation 
test cases for network range 
added member field in Network structure


JkhatriInfobox (approved):
LGTM

PR #293, opened 2025-04-24
test case changes 

Chaithra001:
test case changes  and fixed address code changes


JkhatriInfobox (approved):
LGTM: [QA Verified functionality]

- Changes made for the test cases fix in few modules like DTC

PR #300, opened 2025-10-07
Go version (v1.24.7) upgrade to Fix CVE

JkhatriInfobox:


Aish-sp (approved):

ldez:
The min Go version is the min Go language version: it defines the elements of the syntax that rely on a specific language version.

`infoblox-go-client` is a lib and not an application; bumping the min go version impacts all library consumers, but it fixes nothing inside the lib.

A library should avoid bumping or using patch versions.
And the application should prefer using `toolchain`.

